PR #4, opened 2020-06-30
MGMT-1376 Add repository images list to the assisted-installer.yaml

eranco74:


release/main.py
```diff
@@ -35,20 +39,69 @@ def tag_all(manifest, tag, delete_if_exists=False):
...
+    repos_with_tag = []
+    for repo in manifest_contnet.keys():
+        if gtools.tag_exists(repo, tag):
+            repos_with_tag.append(repo)
+    if repos_with_tag:
+        for repo in repos_with_tag:
+            gtools.delete_tag(repo, tag)
+
+
+def check_images_exists(manifest, tag):
```
ronniel1, line 64: Notice that if we used the '--force' option than this validation is more complicated.
Its not enough to check that the tag exists. Need to verify that the SHA returned for the tag matches the SHA returned for the images with the git-hash tag.

release/main.py
```diff
@@ -35,20 +39,69 @@ def tag_all(manifest, tag, delete_if_exists=False):
...
+                missing_images.append(image)
+    if missing_images:
+        raise Exception("Missinge images %s" % missing_images)
+
+
+def image_exists(image_name, tag):
+    '''Return True if image with the given tag exists in quay.io under ocpmetal namespace'''
+    url = IMAGES_URL_TEMPLATE.format(image_name, tag)
+    response = requests.get(url=url)
+    return response.ok
```
ronniel1, line 86: Should return the SHA of the image (see comment above)

release/main.py
```diff
@@ -35,20 +39,69 @@ def tag_all(manifest, tag, delete_if_exists=False):
...
+        manifest_contnet = yaml.safe_load(manifest_file)
+    logger.info("Deleting %s tag from all repos", tag)
+    gtools = gittools.GitApiUtils()
+    repos_with_tag = []
+    for repo in manifest_contnet.keys():
+        if gtools.tag_exists(repo, tag):
+            repos_with_tag.append(repo)
+    if repos_with_tag:
+        for repo in repos_with_tag:
+            gtools.delete_tag(repo, tag)
```
ronniel1, line 61: Maybe output a message for each tag we delete

release/main.py
```diff
@@ -35,20 +39,69 @@ def tag_all(manifest, tag, delete_if_exists=False):
...
     parser.add_argument('-m', '--manifest', help='Path to manifest file', type=str, default="./assisted-installer.yaml")
     args = parser.parse_args()
-    tag_all(args.manifest, args.tag, args.force)
-
+    if args.delete:
+      untag_all(args.manifest, args.tag)
+    elif args.check:
+        check_images_exists(args.manifest, args.tag)
+    else:
+        tag_all(args.manifest, args.tag, args.force)
```
ronniel1, line 103: Maybe also tag the assisted-installer-deployment repo with the same tag?

release/main.py
eranco74: Thx, didn't think of that.

release/main.py
eranco74: Another PR

PR #9, opened 2020-08-25
update script

oshercc:


release/main.py
```diff
@@ -31,6 +31,8 @@ def tag_all(manifest, tag, delete_if_exists=False):
     gtools = gittools.GitApiUtils()
     repos_with_tag = []
     for repo in manifest_contnet.keys():
+        if not repo.startswith("openshift/"):
+            continue
```
eranco74, line 35: Why skip if it's not openshift?
Add logging

release/gittools.py
```diff
@@ -53,7 +53,7 @@ def delete_tag(self, repo, tag):
         :param str tag: The tag to delete (e.g. 'v1.0.0')
         """
         logging.info('Deleting tag %(tag)s in repository: %(repo)s', dict(tag=tag, repo=repo))
-        ref_url = "%s/%s/git/refs/tags/%s" % (self.GIT_API_REPOS, repo, tag)
+        ref_url = "%s/%s/tags/%s" % (self.GIT_API_REPOS, repo, tag)
```
eranco74, line 56: Sure this is the url?

oshercc: I was also surprised,
but this works and the old isn't working.

release/main.py
oshercc: we don't have permission to tag it, adding logs

ronniel1: @oshercc  I do have permission to tag it
so we should not skip it

PR #13, opened 2020-09-01
Taging tool for stable

oshercc:


tools/update_stable_tags.py
```diff
@@ -0,0 +1,37 @@
...
+#######################################################################
+# This will update the stable (by default) tag to the current revision
+# +  A tag <tag>.%d/%m/%Y-%H-%M will mark the commit
+# notice: * repo has valid push credentials
+#         * working location must be  from in the cloned item
+#######################################################################
+
+retagging_commands = [
+            "git push --delete origin {tag} || true",
+            "git tag -d {tag} || true",
```
ronniel1, line 19: Why are we deleting the previous tag?

tools/update_stable_tags.py
```diff
@@ -0,0 +1,37 @@
...
+ default="stable")
+args = parser.parse_args()
+
+def main():
+    timestamp = datetime.now().strftime("%d/%m/%Y-%H-%M")
+    for cmd in retagging_commands:
+        logging.info("running: {}".format(cmd))
+        subprocess.check_output(cmd.format(tag=args.tag, date=timestamp), shell=True)
+
+if __name__ == "__main__":
```
ronniel1, line 36: Which images does it tag?

tools/update_stable_tags.py
```diff
@@ -0,0 +1,37 @@
...
+retagging_commands = [
+            "git push --delete origin {tag} || true",
+            "git tag -d {tag} || true",
+            "git tag -f {tag}",
+            "git tag -f {tag}.{date}",
+            "git push --tags origin {tag}",
+            ]
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--tag", help="assisted_installer_tag", type=str,
```
ronniel1, line 26: Since in the main usage of the script we'll need to push multiple tags (stable, stable.<data>) maybe it should accept multiple tags arguments and do them at once?

tools/update_stable_tags.py
```diff
@@ -0,0 +1,37 @@
...
+ default="stable")
```
ronniel1, line 27: Don't have default value
if someone runs it by mistake, or forgot to set their tag, we don't really want to push to stable by mistake

PR #26, opened 2020-10-06
MGMT-1247 Assisted-service diff tool

oshercc:


tools/version_docomentaion.py
```diff
@@ -0,0 +1,84 @@
...
+import argparse
+import logging
+import shutil
+import os
+import re
+
+logging.basicConfig(format='%(asctime)s %(message)s')
+logging.getLogger().setLevel(logging.INFO)
+
+REPO_URL = "https://github.com/openshift/assisted-service.git"
```
filanov, line 11: I think that we need it for all the repos, so maybe it should be a parameter?

tools/version_docomentaion.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def process_logs(git_logs_line, version_documentation_list):
+    for line in git_logs_line:
+        line = line.strip().decode("utf-8")
+        if is_line_metadata(line):
+            continue
+        line = line.strip('* ')
+
+        if re.match(valid_commit_regex, line):
```
gamli75, line 50: why are you checking if it's a valid commit regex? we already saw that we have problems when we cherry-pick commits in a branch - please verify that it's working on a branch

tools/version_docomentaion.py
```diff
@@ -0,0 +1,84 @@
...
+        f.write(version_documentation)
+
+
+def is_line_metadata(line):
+    line_starts_with = ['commit', 'Author', 'Date:']
+    for meta_prefix in line_starts_with:
+        if line.startswith(meta_prefix):
+            return True
+    if line == '':
+        return True
```
gamli75, line 80: I think it's should be false

tools/version_docomentaion.py
```diff
@@ -0,0 +1,84 @@
...
+args = parser.parse_args()
+
+
+def main():
+    version_documentation_list = list()
+
+    if not os.path.exists(args.documentation_dir):
+        os.makedirs(args.documentation_dir)
+
+    os.mkdir("temp")
```
gamli75, line 32: why do you need this temp dir?

tools/version_docomentaion.py
```diff
@@ -0,0 +1,84 @@
...
+        line = line.strip('* ')
+
+        if re.match(valid_commit_regex, line):
+            version_documentation_list.append(line)
+
+
+def get_versions_log():
+    subprocess.check_output("git clone {}".format(REPO_URL), shell=True, cwd="temp")
+    raw_log = subprocess.check_output("git log {tagS}...{tagE} ".format(tagS=args.from_version, tagE=args.to_version),
+                                      shell=True, cwd="temp/assisted-service")
```
gamli75, line 57: same remark as @filanov - use repo parameter

tools/version_docomentaion.py
oshercc: nope

tools/version_docomentaion.py
oshercc: we clone the repo to it to get the logs for processing

tools/version_docomentaion.py
oshercc: that's exactly why I take only the validated lines, the metadata is not documented

PR #29, opened 2020-11-01
Added release tickets tool

ronniel1:
Tool that gets the list of tickets from commits between two releases.
It can print a list of all tickets (in various formats) and can also update the BZ tickets with the "fixed_in" field


YuviGold (approved):
LGTM

PR #30, opened 2020-11-16
export triage tickets that closed in the last 7d

ErezAlster:


tools/jira_cmd.py
```diff
@@ -340,8 +340,8 @@ def main(args):
     selectors.add_argument("-nf", "--named-filter", action=buildEpicFilterAction, dest="search_query",
                            help="Search for Epics matching the given named filter")
     selectors.add_argument("-tt", "--triaging-tickets", action='store_const',
-                           dest="search_query", const='project = MGMT AND component = "Assisted-installer Triage"',
-                           help="Search for Assisted Installer triaging tickets")
+                           dest="search_query", const='project = MGMT AND component = "Assisted-installer Triage" AND resolved > -7d',
```
ronniel1, line 343: '-tt' is meant to return all the triaging tickets.

You can either add a new param for this, or just use the '-s' param and specify the query

ErezAlster: ok. missed the a row below with the -rtt
I will change it to use resolved instead of created

PR #43, opened 2021-02-10
Osher/mgmt 3843

oshercc:


tools/new_minor_OCP_version.py
```diff
@@ -243,17 +368,22 @@ def get_latest_ocp_version():
...
+    release_json = get_default_release_json()
 
-    release_image = json.loads(res.text)[ocp_version_major]['release_image']
+    release_image = release_json[ocp_version_major]['release_image']
 
     result = OCP_VERSION_REGEX.search(release_image)
     ai_latest_ocp_version = result.group(1)
     return ai_latest_ocp_version
 
+0
```
omertuc, line 379: 0

tools/new_minor_OCP_version.py
```diff
@@ -120,6 +136,109 @@ def cmd_with_git_ssh_key(key_file):
...
+
+    change_version_in_files(rhcos_version_from_default, rchos_version_from_iso, RCHOS_VERSION_REPLACE_CONTEXT)
+    cmd(["make", "update-ocp-version"], cwd=ASSISTED_SERVICE_CLONE_DIR)
+    branch = commit_and_push_version_update_changes(rhcos_future_latest_release, task)
+
+    github_pr = open_pr(args, rhcos_future_default_release, rhcos_future_latest_release, task)
+    unhold_pr(github_pr)
+
+
+def rhcos_version_update(args):
```
omertuc, line 176: Looks almost identical to rhcos_future_version_update, can they share some code?

Jenkinsfile.ocp_version_update
```diff
@@ -31,13 +31,13 @@ pipeline {
...
-    post {
-        failure {
-            script {
-                def data = [text: "OCP version failed, see: ${BUILD_URL}"]
-                writeJSON(file: 'data.txt', json: data, pretty: 4)
-            }
-            sh '''curl -X POST -H 'Content-type: application/json' --data-binary "@data.txt"  https://hooks.slack.com/services/$SLACK_TOKEN'''
-        }
-    }
+    //post {
```
omertuc, line 34: TODO: Uncomment when done testing

tools/new_minor_OCP_version.py
```diff
@@ -120,6 +136,109 @@ def cmd_with_git_ssh_key(key_file):
...
+    cmd(["make", "update-ocp-version"], cwd=ASSISTED_SERVICE_CLONE_DIR)
+    branch = commit_and_push_version_update_changes(rhcos_latest_release, task)
+
+    github_pr = open_pr(args, rhcos_default_release, rhcos_latest_release, task)
+    unhold_pr(github_pr)
+
+def get_rchos_version_from_iso(rhcos_latest_release):
+    live_iso_url = RCHOS_LATEST_LIVE_ISO_URL.format(version=rhcos_latest_release)
+    with tempfile.NamedTemporaryFile() as tmp_live_iso_file:
+        subprocess.check_output(
```
omertuc, line 212: As long as you're inside the `with` statement, the file is opened, which might cause issues when trying to overwrite it with curl. Instead, use the `with` statement just to capture the name of the file, then do everything outside the scope of the `with` statement. Similar to this:
https://github.com/openshift/assisted-test-infra/blob/2b3aec7af8559eace8f9d151e7ab78f6536e8aa5/discovery-infra/test_infra/controllers/node_controllers/libvirt_controller.py#L209-L212

omertuc, line 212: Use `cmd` - it logs every call to it which is useful for debugging issues. Also what's the point of using `shell` here?

tools/new_minor_OCP_version.py
```diff
@@ -120,6 +136,109 @@ def cmd_with_git_ssh_key(key_file):
...
+            DOWNLOAD_LIVE_ISO_CMD.format(live_iso_url=live_iso_url, out_file=tmp_live_iso_file.name), shell=True)
+        try:
+            os.remove("tmp/zipl.prm")
+        except:
```
omertuc, line 216: Don't use broad except, at-the very least use `except Exception` to avoid capturing exceptions caused by kill / interrupt signals that might come from a user trying to stop the job

tools/new_minor_OCP_version.py
```diff
@@ -120,6 +136,109 @@ def cmd_with_git_ssh_key(key_file):
...
+            pass
+        subprocess.check_output(f"7z x {tmp_live_iso_file.name} zipl.prm", shell=True, cwd="/tmp")
```
omertuc, line 218: use cmd, no need for shell

omertuc:
Any progress on this PR? The GitLab code is broken and I don't want to make any changes on master while this PR is still open, too many conflicts.

Error that just happened in Jenkins:
http://assisted-jenkins.usersys.redhat.com/job/ocp_version_update/80/console

PR #52, opened 2021-03-16
Added tool for creating 'triage' tickets for CI tests that are in 'Fa…

ronniel1:
…iling' status

### DIFF
--- a/tools/create_testgrid_tickets.py
+++ b/tools/create_testgrid_tickets.py
@@ -1,0 +1,161 @@
+#!/usr/bin/env python3
+"""
+This script gets a list of the filed tests from testgrid
+For each failure, which does not already has an open triaging Jira ticket, it creates one
+"""
+
+
+import argparse
+import logging
+import netrc
+import os
+import sys
+import re
+from urllib.parse import urlparse
+import requests
+import jira
+
+
+DEFAULT_WATCHERS = ["ronniela", "odepaz"]
+
+
+TEST_VERSION_REGEX = re.compile(r'.*-(\d\.[^-]+)-.*')
+TEST_STATUS_URL = "https://testgrid.k8s.io/redhat-assisted-installer/summary"
+TEST_RUNS_URL = 'https://testgrid.k8s.io/redhat-assisted-installer/table?tab={}&width=10&dashboard=redhat-assisted-installer'
+JIRA_SERVER = "https://issues.redhat.com/"
+DEFAULT_NETRC_FILE = "~/.netrc"
+JIRA_SUMMARY = "CI test: {test_id} ({last_pass})"
+
+JIRA_DESCRIPTION = """
+{{color:red}}Do not manually edit this description, it will get automatically over-written{{color}}
+
+*Test Status*: {status}
+*Test Details:* [{test_id}|https://testgrid.k8s.io/redhat-assisted-installer#{test_id}]
+*Last Test failure:* https://prow.ci.openshift.org/view/gcs/origin-ci-test/logs/{test_id}/{last_fail}
+{tests_list}
+"""
+
+
+def get_credentials_from_netrc(server, netrc_file=DEFAULT_NETRC_FILE):
+    cred = netrc.netrc(os.path.expanduser(netrc_file))
+    username, _, password = cred.authenticators(server)
+    return username, password
+
+
+def get_jira_client(username, password):
+    logger.info("log-in with username: %s", username)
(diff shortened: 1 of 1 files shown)


tools/create_testgrid_tickets.py
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env python3
+
+# This script gets a list of the filed tests from testgrid
+# For each failure, which does not already has an open triaging Jira ticket, it creates one
+#
```
osherdp, line 5: Replace with module docstring (triple quotes)

tools/create_testgrid_tickets.py
```diff
@@ -0,0 +1,161 @@
...
+{{color:red}}Do not manually edit this description, it will get automatically over-written{{color}}
+
+*Test Status*: {status}
+*Test Details:* [{test_id}|https://testgrid.k8s.io/redhat-assisted-installer#{test_id}]
+*Last Test failure:* https://prow.ci.openshift.org/view/gcs/origin-ci-test/logs/{test_id}/{last_fail}
+{tests_list}
+"""
+
+
+def get_credentials_from_netrc(server, netrc_file=DEFAULT_NETRC_FILE):
```
osherdp, line 38: maybe we should have it in some utility module

tools/create_testgrid_tickets.py
```diff
@@ -0,0 +1,161 @@
...
+    return [i.fields.summary for i in jclient.search_issues(query, maxResults=None)]
+
+def add_watchers(jclient, issue):
+    for watcher in DEFAULT_WATCHERS:
+        jclient.add_watcher(issue.key, watcher)
+
+def get_last_failure_instance(test_id):
+    res = requests.get(TEST_RUNS_URL.format(test_id))
+    res.raise_for_status()
+    j = res.json()
```
osherdp, line 63: just inline variable ``j``

tools/create_testgrid_tickets.py
```diff
@@ -0,0 +1,161 @@
...
+    test_status = test_data['status']
+    i = get_last_failure_instance(test_id)
+    new_issue = jclient.create_issue(project="MGMT",
+                                     description=format_description({"test_id": test_id,
+                                                                     "last_fail": i,
+                                                                     "failed_tests": failed_tests,
+                                                                     "status": test_status,
+                                                                    }),
+                                     summary=summary,
+                                     versions=[{'name': 'OpenShift {}'.format(TEST_VERSION_REGEX.findall(test_id)[0])}],
```
osherdp, line 100: If we either way only want one result:
```suggestion
                                     versions=[{'name': 'OpenShift {}'.format(TEST_VERSION_REGEX.search(test_id))}],
```

tools/create_testgrid_tickets.py
```diff
@@ -0,0 +1,161 @@
...
+    for failed_test in failed_tests:
+        create_jira_ticket(jclient, existing_tickets, failed_test, failed_tests[failed_test])
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    loginGroup = parser.add_argument_group(title="login options")
+    loginArgs = loginGroup.add_mutually_exclusive_group()
+    loginArgs.add_argument("--netrc", default="~/.netrc", required=False, help="netrc file")
+    loginArgs.add_argument("-up", "--user-password", required=False,
```
osherdp, line 149: maybe rename to ``--auth`` ?

osherdp (changes requested):

tools/create_testgrid_tickets.py
ronniel1: done

tools/create_testgrid_tickets.py
ronniel1: You are right.
This is a refactor that needs to be done on the entire repo. There are many tools that duplicate this code.
This will be handled as part of https://issues.redhat.com/browse/MGMT-3945

tools/create_testgrid_tickets.py
ronniel1: done

tools/create_testgrid_tickets.py
ronniel1: This does not work cause search returns an object

tools/create_testgrid_tickets.py
ronniel1: This is alligned with the other similar script.
I guess that once we consolidate all the jira related code, we can think on the parameter name

osherdp (approved):

PR #56, opened 2021-03-21
triage_stats tool

ronniel1:
Added triage_stats tool which outputs triage tickets stats, grouped by weeks, for the past few weeks

### DIFF
--- a/tools/jira_cmd.py
+++ b/tools/jira_cmd.py
@@ -4,6 +4,7 @@
 # A tool to perform various operations on Jira
 
 import os
+import io
 import sys
 import re
 import csv
@@ -16,6 +17,7 @@ from urllib.parse import urlparse
 from collections import Counter
 from tabulate import tabulate
 import jira
+import json
 
 
 JIRA_SERVER = "https://issues.redhat.com/"
@@ -24,6 +26,7 @@ FIELD_SPRINT = 'customfield_12310940'
 FIELD_CONTRIBUTORS = 'customfield_12315950'
 
 MAX_RESULTS = 100
+TRIAGE_MAX_RESULTS = 300
 
 SEARCH_QUERY_EPICS = 'project = MGMT AND component = "MGMT OCP Metal" AND type = Epic AND filter = '
 CURRENT_VERSION_FILTER = '12347072'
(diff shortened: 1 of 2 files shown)


tools/jira_cmd.py
```diff
@@ -118,9 +120,12 @@ def get_data_for_print(issues, issues_count=None, print_fields=None):
         if 'component' in headers:
             row['component'] = [c.name for c in i.fields.components]
         if 'priority' in headers:
-            row['priority'] = i.fields.priority.name
+            try:
+                row['priority'] = i.fields.priority.name
+            except Exception:
+                row['priority'] = i.fields.priority
```
osherdp, line 126: What happens here? and can we make it simpler?

tools/jira_cmd.py
```diff
@@ -521,6 +535,9 @@ def handle_sprint_update(args, jiraTool, issues):
 
 def main(args):
     j = jira_netrc_login(args.netrc)
+    max_results = args.max_results
+    if args.max_results == MAX_RESULTS and args.linked_issues:
```
osherdp, line 539: Maybe let's add a warning log to indicate of this limit?

tools/triage_stats.py
```diff
@@ -0,0 +1,44 @@
...
+
+class TriageStats:
+    def __init__(self):
+        parser = jira_cmd.build_parser()
+        def my_exit(msg):
+            pass
+        parser.error = my_exit
+        self.args = parser.parse_args()
+        self.args.print_json = True
+        self.args.linked_issues = True
```
osherdp, line 20: I don't get it
Why we need to call it with the cli parser? Can't we call a regular function?

tools/triage_stats.py
```diff
@@ -0,0 +1,44 @@
...
+        print(self.args.search_query)
+        j = jira_cmd.main(self.args)
+        data = json.loads(j)
+        for i in data:
+            if i['key'] not in self.stats:
+                self.stats[i['key']] = dict(key=i['key'], summary=i['summary'], status=i['status'])
+            self.stats[i['key']]['week -{}'.format(past_week_num)] = i['count']
+
+
+a = TriageStats()
```
osherdp, line 37: Let's guard it inside ``if __name__ == "__main__":``

osherdp (changes requested):

tools/jira_cmd.py
ronniel1: In some projects the priority field is an object that has a name property and in others it is not such an object

tools/jira_cmd.py
ronniel1: ok

tools/triage_stats.py
ronniel1: Since most of the code get the parser struct, we need to build this struct and populate it

tools/triage_stats.py
ronniel1: ok

osherdp (approved):

PR #57, opened 2021-03-21
MGMT-4684 triage signatures: added ConsoleTimeoutSignature

danielerez:
Added ConsoleTimeoutSignature for detecting 'waiting for console' error.
Specifically, OCP 4.7 and VMware vms hosts.
See: https://bugzilla.redhat.com/1926345

Expected signature:
```
h1. Console timeout - VMware hosts / OCP 4.7
h2. Waiting for console timeout probably due to VMware hosts on OCP 4.7
See: https://bugzilla.redhat.com/1926345
```

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -532,13 +532,51 @@ class IOErrorsSignature(Signature):
                 self._update_triaging_ticket(issue_key, report, should_update=should_update)
 
 
+class ConsoleTimeoutSignature(Signature):
+    """
+    This signature looks for 'waiting for console' error in cluster's status_info.
+    If OpenShift version is 4.7 and hosts are VMware vms, outputs a warning.
+    Due to: https://bugzilla.redhat.com/1926345
+    """
+
+    def __init__(self, jira_client):
+        super().__init__(jira_client, comment_identifying_string="h1. Console timeout - VMware hosts / OCP 4.7")
+
+    def _update_ticket(self, url, issue_key, should_update=False):
+        url = self._logs_url_to_api(url)
+
+        try:
+            md = get_metadata_json(url)
+        except Exception as e:
+            logger.error("Error getting metadata for %s at %s: %s, it may have been deleted", issue_key, url, e)
+            return
+
+        def isVMware(host):
+            inventory = json.loads(host['inventory'])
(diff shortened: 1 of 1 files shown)


danielerez:
@YuviGold @osherdp

tools/add_triage_signature.py
```diff
@@ -532,13 +532,51 @@ def _update_ticket(self, url, issue_key, should_update=False):
                 self._update_triaging_ticket(issue_key, report, should_update=should_update)
 
 
+class ConsoleTimeoutSignature(Signature):
+    '''
```
osherdp, line 536: replace with triple double-quotes (" x 3)

osherdp (approved):

tools/add_triage_signature.py
```diff
@@ -532,13 +532,51 @@ def _update_ticket(self, url, issue_key, should_update=False):
...
+        cluster = md['cluster']
+        status_info = cluster['status_info']
+        openshift_version = cluster['openshift_version']
+        report = ""
+        if "waiting for console" in status_info and "4.7" in openshift_version:
+            for host in cluster['hosts']:
+                inventory = json.loads(host['inventory'])
+                product_name = inventory['system_vendor'].get('product_name')
+                if "VMware" not in product_name:
+                    return
```
YuviGold, line 563: In case the first host is not a vmware, you will exit the function.
You wanted to use `continue`?

danielerez: I actually wanted to add the signature only if all hosts are vmware, to imitate the exact scenario.
do you think it should be added if any are vmware just in case?

osherdp: yes, let's do that just in case
even though 99% of the time people use the same platform across all machines

tools/add_triage_signature.py
danielerez: done. 
both are similar: https://www.digitalocean.com/community/tutorials/how-to-format-text-in-python-3#multiple-lines
though for consistency, it's indeed recommended per PEP 257: https://www.python.org/dev/peps/pep-0257/

tools/add_triage_signature.py
danielerez: done

PR #58, opened 2021-03-22
Fixed triage_stats to avoid passing arguments to the underlying jira_cmd

ronniel1:


YuviGold:
If eventually it calls 
```python
jira_cmd.main(self.args)
```
shouldn't we want to use the same arg parser? (So `--help` would work)

ronniel1:
> If eventually it calls
> 
> ```python
> jira_cmd.main(self.args)
> ```
> 
> shouldn't we want to use the same arg parser? (So `--help` would work)

No, this script does not pass user args to underlying jira_cmd tool.
for jira_cmd to work, it needs to build the jira_cmd argparse structure and populate it. That is why it calls parse_args() of jira_cmd without any arguments

PR #59, opened 2021-03-22
MGMT-4861 Use nestedarchive library to more simply access logs

omertuc:
Use nestedarchive library to more simply access logs

The log cache contains nestedarchive.RemoteNestedArchive objects that
take care of downloading, extracting and providing easy cached access
to log files, even in nested log archives.

See pypi.org/project/nestedarchive

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -9,11 +9,9 @@ import netrc
 import os
 import re
 import sys
-import tarfile
 import tempfile
 from collections import OrderedDict, defaultdict
 from datetime import datetime
-from tempfile import NamedTemporaryFile, TemporaryDirectory
 from textwrap import dedent
 from typing import List
 from urllib.parse import urlparse
@@ -21,6 +19,7 @@ from urllib.parse import urlparse
 import colorlog
 import dateutil.parser
 import jira
+import nestedarchive
 import requests
 import tqdm
 from fuzzywuzzy import fuzz
@@ -92,12 +91,22 @@ def get_events_json(cluster_url, cluster_id):
     return res.json()
 
 
-def get_cluster_logs(cluster_url: str, cluster_id: str, output_file: str):
-    res = requests.get(f"{cluster_url}/cluster_{cluster_id}_logs.tar")
-    res.raise_for_status()
+@functools.lru_cache(maxsize=100)
+def get_remote_archive(tar_url: str) -> nestedarchive.RemoteNestedArchive:
+    return nestedarchive.RemoteNestedArchive(tar_url, init_download=True)
 
-    with open(output_file, "wb") as f:
-        f.write(res.content)
+
+def get_triage_logs_tar(triage_url: str, cluster_id: str) -> nestedarchive.RemoteNestedArchive:
+    tar_url = f"{triage_url}/cluster_{cluster_id}_logs.tar"
+    return get_remote_archive(tar_url)
+
+
+def get_host_log_file(triage_logs_tar: nestedarchive.RemoteNestedArchive,
+                      host_id: str, filename: str):
+    # The file is already uniquely determined by the host_id, we can omit the hostname
+    hostname = "*"
+
+    return triage_logs_tar.get(f"{hostname}.tar.gz/logs_host_{host_id}/{filename}")
 
 
 ############################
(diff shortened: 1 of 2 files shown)


tools/add_triage_signature.py
```diff
@@ -683,6 +662,10 @@ def get_issues(jclient, issue, only_recent):
     return get_all_triage_tickets(jclient, only_recent=only_recent)
 
 
+def browse(issue_key: str):
```
osherdp, line 665: we need a more indicative name than ``browse``

tools/add_triage_signature.py
```diff
@@ -494,42 +497,36 @@ def _update_ticket(self, url, issue_key, should_update=False):
...
+                dmesg_file = get_host_log_file(log_cache=self.log_cache,
+                    cluster_url=url, host_id=host_id, cluster_id=cluster_id,
+                    filename="dmesg.logs")
+            except FileNotFoundError:
+                # Host doesn't have dmesg
+                continue
+            except requests.exceptions.HTTPError as e:
+                if e.errno == 404:
+                    logger.warning(f"{browse(issue_key)} doesn't have a log tar, skipping")
+                continue
```
osherdp, line 515: maybe we should at least log the error?

tools/add_triage_signature.py
```diff
@@ -91,12 +91,14 @@ def get_events_json(cluster_url, cluster_id):
...
 
-def get_cluster_logs(cluster_url: str, cluster_id: str, output_file: str):
-    res = requests.get(f"{cluster_url}/cluster_{cluster_id}_logs.tar")
-    res.raise_for_status()
+def get_host_log_file(log_cache: LRU, cluster_url: str, cluster_id: str, host_id: str, filename: str):
+    tar_url = f"{cluster_url}/cluster_{cluster_id}_logs.tar"
+    remote_nested_tar: nestedarchive.RemoteNestedArchive = log_cache[tar_url]
+
+    # The file is already uniquely determined by the host_id, we can omit the hostname
+    hostname = "*"
```
osherdp, line 99: just inline this variable, I don't think it's that important

osherdp (changes requested):

tools/add_triage_signature.py
omertuc: Wanted to add a comment explaining why I do that, if I inline it - there won't be a good place to put that comment. Also just putting `*` doesn't make it really clear that it's the hostname unless you're really familiar with and remember the structure of the logs

tools/add_triage_signature.py
omertuc: Forgot to raise in case the error is not 404. But on 404 I intentionally don't log the error because it's really spammy

tools/add_triage_signature.py
omertuc: sure

tools/add_triage_signature.py
```diff
@@ -745,9 +728,14 @@ def add_signatures(jclient, url, issue_key, should_update=False, signatures=None
         should_update = True
         signatures_to_add = [v for k, v in name_to_signature.items() if k in signatures]
 
+    # boltons LRU is practically a size-limited defaultdict where the least recently used keys get evicted automatically
```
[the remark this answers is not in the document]
omertuc: I guess I could replace the dict with a function that receives the key as input and the returns the object. Then decorate it with `lru_cache`. It'll "save" me from having to pass the log cache to all the signatures because it kinda turns the dict into a global variable that masquerades as a function.

Will do that

YuviGold:
Very nice! Please rebase

YuviGold:
Could you show how much it improves performance?

omertuc:
> Could you show how much it improves performance?

The main goal was to simplify how the logs are accessed, not just improve performance.

There are no performance benefits currently gained from this PR, since the caching only makes a difference when there are multiple signatures accessing tarred logs - those signatures will use the same downloaded / extracted tars behind the scenes, so it'll save them from re-downloading and re-extracting archives of the same ticket.

tools/add_triage_signature.py
```diff
@@ -92,12 +91,22 @@ def get_events_json(cluster_url, cluster_id):
     return res.json()
 
 
-def get_cluster_logs(cluster_url: str, cluster_id: str, output_file: str):
-    res = requests.get(f"{cluster_url}/cluster_{cluster_id}_logs.tar")
-    res.raise_for_status()
+@functools.lru_cache(maxsize=100)
+def get_remote_archive(tar_url: str) -> nestedarchive.RemoteNestedArchive:
```
ronniel1, line 95: I think using annotation, with out having something actually verify them is bad
They are like remarks.
They will be misleading the next time someone changes the code.

omertuc: Removed all type annotations in d82a042

YuviGold: I have to disagree with you on that.
For example, 
```
- def group_similar_strings(ls: List[str], ratio: int) -> List[str]:
+ def group_similar_strings(ls, ratio):
```
I can't really know how the grouping works unless I would investigate the function logic. It could be concatenated into a single string. The `List` indication immediately guides me on how the function works.

omertuc: I don't think the benefits of type hints are in question, I believe everyone agrees on that. @ronniel1 prefers we have a discussion about whether we want them or not before introducing them in the code because they have the potential to become confusing as they grow out of sync with reality because there's nothing to keep them in check.

PR #61, opened 2021-03-30
use a more indicative comment for vmware bug

osherdp:
This references the right BZ ticket (not the duplication), and putting the triage root-cause issue directly.
/cc @YuviGold

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -476,7 +476,7 @@ class LibvirtRebootFlagSignature(Signature):
             self._update_triaging_ticket(issue_key, report, should_update=should_update)
 
 
-class IOErrorsSignature(Signature):
+class MediaDisconnectionSignature(Signature):
     '''
     The signature downloads cluster logs, and go over all the hosts logs files searching for a 'dmesg' file
     Then, it looks for i/o errors (disks, media) patterns and groups them per host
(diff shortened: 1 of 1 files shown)


tools/add_triage_signature.py
```diff
@@ -569,10 +569,14 @@ def isVMware(host):
...
-        if "waiting for console" in status_info and "4.7" in openshift_version:
-            if any(isVMware(host) for host in cluster['hosts']):
-                report = "h2. Waiting for console timeout probably due to VMware hosts on OCP 4.7\n"
-                report += "See: https://bugzilla.redhat.com/1926345\n"
+        if ("waiting for console" in status_info and
+                "4.7" in openshift_version and
+                any(isVMware(host) for host in cluster['hosts'])):
+            report = "\n".join([
+                "h2. Waiting for console timeout probably due to VMware hosts on OCP 4.7",
+                "You can mark it as caused by issue [MGMT-4454|https://issues.redhat.com/browse/MGMT-4454]",
```
YuviGold, line 577: Can we set the issue link automatically as well? :)

osherdp: Probably we would do that next
I think we'll discuss it in the next triage weekly. Each signature have its own level of strength, such that some should just be informing, some can be connected automatically to tickets, some can just close the ticket, etc.

PR #63, opened 2021-04-01
MGMT-4841 - Add signature for agent step failures

omertuc:
Every time an agent step fails, it logs the exit code, stdout and
stderr of the failed step. It's usually interesting to see those
failures, so this commit adds a triage signature that scans the
agent logs for those failure logs and includes them in a table

Also simlified exception handling inside signatures.

Example:
image

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -77,11 +77,18 @@ def days_ago(datestr):
         return 9999
 
 
+class FailedToGetMetadataException(Exception):
+    pass
+
+
 @functools.lru_cache(maxsize=1000)
 def get_metadata_json(cluster_url):
-    res = requests.get("{}/metdata.json".format(cluster_url))
-    res.raise_for_status()
-    return res.json()
+    try:
+        res = requests.get("{}/metdata.json".format(cluster_url))
+        res.raise_for_status()
+        return res.json()
+    except Exception as e:
+        raise FailedToGetMetadataException from e
 
 
 @functools.lru_cache(maxsize=1000)
@@ -96,9 +103,16 @@ def get_remote_archive(tar_url):
     return nestedarchive.RemoteNestedArchive(tar_url, init_download=True)
 
 
+class FailedToGetLogsTarException(Exception):
+    pass
+
+
 def get_triage_logs_tar(triage_url, cluster_id):
-    tar_url = f"{triage_url}/cluster_{cluster_id}_logs.tar"
-    return get_remote_archive(tar_url)
+    try:
+        tar_url = f"{triage_url}/cluster_{cluster_id}_logs.tar"
+        return get_remote_archive(tar_url)
+    except requests.exceptions.HTTPError as e:
+        raise FailedToGetLogsTarException from e
 
 
 def get_host_log_file(triage_logs_tar, host_id, filename):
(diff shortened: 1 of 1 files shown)


tools/add_triage_signature.py
```diff
@@ -582,14 +564,82 @@ def isVMware(host):
             self._update_triaging_ticket(issue_key, report, should_update=should_update)
 
 
+class AgentStepFailureSignature(Signature):
+    """
+    This signature creates a report of all agent step failures that were encountered in the logs
+    """
+    PATTERN = re.compile(
```
omertuc, line 571: It looks less worse with syntax highlighting:
image

tools/add_triage_signature.py
```diff
@@ -77,14 +77,20 @@ def days_ago(datestr):
...
 @functools.lru_cache(maxsize=1000)
 def get_metadata_json(cluster_url):
-    res = requests.get("{}/metdata.json".format(cluster_url))
-    res.raise_for_status()
-    return res.json()
+    try:
+        res = requests.get("{}/metdata.json".format(cluster_url))
+        res.raise_for_status()
+        return res.json()
+    except Exception as e:
```
osherdp, line 90: maybe only on 404?

tools/add_triage_signature.py
```diff
@@ -96,9 +102,16 @@ def get_remote_archive(tar_url):
...
+    pass
+
+
 def get_triage_logs_tar(triage_url, cluster_id):
-    tar_url = f"{triage_url}/cluster_{cluster_id}_logs.tar"
-    return get_remote_archive(tar_url)
+    try:
+        tar_url = f"{triage_url}/cluster_{cluster_id}_logs.tar"
+        return get_remote_archive(tar_url)
+    except requests.exceptions.HTTPError as e:
```
osherdp, line 113: same as above. 5xx should be noticeable

YuviGold:
Could you post how the signature looks like?

tools/add_triage_signature.py
osherdp: yes, I'm not sure how to make it more readable
maybe separate it into a couple of regexps, as time, level and msg belong to the top-level message and msg is composed by several fields

osherdp (changes requested):

tools/add_triage_signature.py
```diff
@@ -77,14 +77,20 @@ def days_ago(datestr):
...
-    return res.json()
+    try:
+        res = requests.get("{}/metdata.json".format(cluster_url))
+        res.raise_for_status()
+        return res.json()
+    except Exception as e:
+        raise FailedToGetMetadataException from e
 
 
-@functools.lru_cache(maxsize=1000)
```
YuviGold, line 87: Why was it removed?

tools/add_triage_signature.py
```diff
@@ -582,14 +564,81 @@ def isVMware(host):
             self._update_triaging_ticket(issue_key, report, should_update=should_update)
 
 
+class AgentStepFailureSignature(Signature):
+    """
+    This signature creates a report of all agent step failures that were encountered in the logs
+    """
+    PATTERN = re.compile(
```
YuviGold, line 571: Can you set a link to the place this log happens? It seems that we will have to be consistent with it

tools/add_triage_signature.py
omertuc: Already had this discussion in a different PR. I didn't add this `except` clause, I just moved it from the signatures to this function. Actually changing it would cause behavior change that is out of the scope of this PR.

tools/add_triage_signature.py
omertuc: Don't want behavior changes, I only moved it from the signatures to here to simplify the code. Can change this in a different PR

tools/add_triage_signature.py
omertuc: Whoops

ronniel1 (approved):

PR #65, opened 2021-04-07
signature: format std output as code

danielerez:
To avoid "\n" in std and error outputs, split text into lines and format each as code.

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -584,7 +584,8 @@ class AgentStepFailureSignature(Signature):
     )
 
     MAX_FAILURES_PER_HOST = 10
-    TRUNCATE_OUTPUT_LENGTH = 1500
+    MAX_LINE_LENGTH = 1000
+    TRUNCATE_OUTPUT_LINES = 100
 
     def __init__(self, jira_client):
         super().__init__(jira_client, comment_identifying_string="h1. Agent step failures")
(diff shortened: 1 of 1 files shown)


tools/add_triage_signature.py
```diff
@@ -594,11 +594,18 @@ def _prepare_output(cls, output):
         if len(output) == 0:
             return output
 
-        if len(output) > cls.TRUNCATE_OUTPUT_LENGTH:
-            half = cls.TRUNCATE_OUTPUT_LENGTH // 2
-            output = f"{output[:half]} ... **OUTPUT HAS BEEN TRUNCATED** ... {output[-half:]}"
+        pre = "{code:borderStyle=none|bgColor=transparent} "
+        post = " {code}"
+        output_list = output.replace("|", "¦").split("\\n")
```
omertuc, line 599: output_lines

tools/add_triage_signature.py
```diff
@@ -594,11 +594,18 @@ def _prepare_output(cls, output):
...
 
-        return "{code:none}%s{code}" % (output.replace("|", "¦"))
+        if len(output_list) > cls.TRUNCATE_OUTPUT_LINES:
+            half = cls.TRUNCATE_OUTPUT_LINES // 2
+            output_list = output_list[:half] + ["**OUTPUT HAS BEEN TRUNCATED**"] + output_list[-half:]
+
+        output_list = list(filter(lambda x: x != "\\\"", output_list))
```
omertuc, line 605: nit: `list(filter(lambda x:` can be expressed more "elegantly" with a list comprehension `if` filter:

`[line for line in output_lines if line != "\\\""]`

omertuc:
Did you test this in the browser? IIRC, I tried to do something similar and JIRA just messed up my entire table...

tools/add_triage_signature.py
```diff
@@ -594,11 +594,18 @@ def _prepare_output(cls, output):
...
+        output_list_wrapped = [pre + sub + post for sub in output_list]
```
osherdp, line 606: what ``sub`` means? maybe rename

tools/add_triage_signature.py
```diff
@@ -584,7 +584,7 @@ class AgentStepFailureSignature(Signature):
     )
 
     MAX_FAILURES_PER_HOST = 10
-    TRUNCATE_OUTPUT_LENGTH = 1500
+    TRUNCATE_OUTPUT_LINES = 100
```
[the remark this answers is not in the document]
danielerez: How much extreme length? :) It seems line wrap is supported, for moderate length at least.

tools/add_triage_signature.py
danielerez: sure

osherdp (changes requested):

tools/add_triage_signature.py
danielerez: I actually think the lambda syntax is nicer, but just matter of style I guess.

tools/add_triage_signature.py
danielerez: sublist :) I'll change to 'line', maybe clearer.

tools/add_triage_signature.py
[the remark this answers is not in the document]
omertuc: The reason I added the truncate in the first place because without truncating, the long line length of the install command froze my browser. I guess 1000 could be a good limit

danielerez:
> Did you test this in the browser? IIRC, I tried to do something similar and JIRA just messed up my entire table... Can you attach screenshots of what it looks like?

Yeah tried it with a JIRA comment. A bit too much padding/margin between lines, but that cause there's a separate panel for each one, couldn't find a good way to modify it.

tools/add_triage_signature.py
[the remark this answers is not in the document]
danielerez: Oh this line is now split into multiple lines, so should be fine. But I can add some limitation just in case.

danielerez:
ping? :)

PR #67, opened 2021-04-08
MGMT-4934 - Timeout signature now less relevant

omertuc:
Maybe we should wait a couple of hours before merging this, to wait for all clusters that have already started installation to finish.

After a few weeks we should remove the signature altogether - once we see that bug has been truly fixed and has stopped causing console timeouts.


omertuc:
I just realized this will mess up older tickets affected by this bug. 

/hold until I add a date `if` statement

osherdp (approved):

PR #68, opened 2021-04-11
MGMT-3439 - Add S.M.A.R.T. information to storage triage signature

omertuc:
@oshercc @ronniel1

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -15,6 +15,7 @@ from datetime import datetime
 from textwrap import dedent
 from typing import List
 from urllib.parse import urlparse
+from enum import Flag, auto
 
 import colorlog
 import dateutil.parser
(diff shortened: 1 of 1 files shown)


tools/add_triage_signature.py
```diff
@@ -383,8 +384,87 @@ class StorageDetailSignature(Signature):
     def __init__(self, jira_client):
         super().__init__(jira_client, comment_identifying_string="h1. Host storage details:")
 
-    def _update_ticket(self, url, issue_key, should_update=False):
+    class SmartctlExitCode(Flag):
+        """
+        Taken from `man smartctl` "EXIT STATUS" section
+        """
```
osherdp, line 390: Single line

tools/add_triage_signature.py
```diff
@@ -383,8 +384,87 @@ class StorageDetailSignature(Signature):
...
+        exit_code = smartctl["exit_status"]
+
+        if exit_code != 0:
+            output.append("non-zero smartctl exit code: " + ", ".join(flag.name for flag in cls.SmartctlExitCode
+                                                                      if flag in cls.SmartctlExitCode(exit_code)))
+            if cls.SmartctlExitCode(exit_code) == cls.SmartctlExitCode.SOME_ATA_COMMAND_FAILED_OR_SMART_CHECKSUM_ERROR:
+                if "smart_status" not in smart:
+                    output.append(
+                        "{color:green}*this typically happens on virtual disks*{color} that don't actually have S.M.A.R.T. information,"
+                        " but may also happen due to other reasons")
```
osherdp, line 413: So I wonder if green is the right color in this case
Maybe no color is the right thing here

tools/add_triage_signature.py
```diff
@@ -383,8 +384,87 @@ class StorageDetailSignature(Signature):
...
+
+        if "messages" in smartctl:
+            for message in smartctl["messages"]:
```
osherdp, line 416:
```
for message in smartctl.get("messages", []):
```

tools/add_triage_signature.py
osherdp: or some other neutral color

tools/add_triage_signature.py
```diff
@@ -383,8 +384,87 @@ class StorageDetailSignature(Signature):
...
+
+        if "power_on_time" in smart:
+            if "hours" in smart["power_on_time"]:
+                hours = smart["power_on_time"]["hours"]
+                output.append(
+                    f"disk has been powered on for {hours // 24} days and {hours % 24} hours throughout its life")
+
+        if "ata_smart_attributes" in smart:
+            table = smart["ata_smart_attributes"].get("table", [])
+            interesting = ("Program_Fail_Count", "Erase_Fail_Count", "Offline_Uncorrectable")
```
osherdp, line 443: maybe rename to ``relevant_attributes``

tools/add_triage_signature.py
```diff
@@ -383,8 +384,87 @@ class StorageDetailSignature(Signature):
...
+        return ", ".join(output)
+
+    @classmethod
+    def _parse_smart(cls, smart_json):
+        try:
+            return cls._parse_smart_internal(json.loads(smart_json))
+        except json.JSONDecodeError as e:
+            return f"Failed to decode S.M.A.R.T. JSON, error: {e}"
+        except Exception as e:
+            return f"S.M.A.R.T. JSON has unexpected format, error: {e}"
```
osherdp, line 465: above exception should be JSON decode error, no?

tools/add_triage_signature.py
```diff
@@ -396,19 +476,24 @@ def _update_ticket(self, url, issue_key, should_update=False):
...
             hosts.append(OrderedDict(
                 id=host['id'],
                 hostname=self._get_hostname(host),
                 disk_name="\n".join(disks_details['name']),
                 disk_type="\n".join(disks_details['type']),
                 disk_path="\n".join(disks_details['path']),
                 disk_bootable="\n".join(disks_details['bootable']),
-                disk_by_path="\n".join(disks_details['by-path'])
+                disk_by_path="\n".join(disks_details['by-path']),
+                SMART="\n".join(disks_details['smart']),
```
osherdp, line 496: even though it's an acronym, I think it should stay ``smart`` because of Python conventions

osherdp (changes requested):

tools/add_triage_signature.py
omertuc: It's a long sentence, and it generally means "Everything is probably OK, this is probably just a virtual disk, SMART is uninteresting". I want to draw the user to the "this is okay part" to reduce mental effort when going over this long table of disks.

tools/add_triage_signature.py
omertuc: Meet you halfway `interesting_attributes`

tools/add_triage_signature.py
omertuc: No, the second message I meant that the structure of the JSON is unexpected, not that the JSON encoding itself is wrong.

tools/add_triage_signature.py
omertuc: This kwarg is directly used as the header of the table shown to the user, and just lowercase "smart" is confusing - not everybody knows this acronym

tools/add_triage_signature.py
omertuc: Improved, somewhat

tools/add_triage_signature.py
osherdp: So change to ``"SMART": "value"``` (other values as well. Although it's weird we show ``disk_name`` and not ``Disk Name`` in a human readable table)

omertuc: Not sure I understand

osherdp:
```suggestion
                "Disk Path": "\n".join(disks_details['path']),
                ...
                "SMART": "\n".join(disks_details['smart']),
```

osherdp (approved):

PR #69, opened 2021-04-18
MGMT-5241 - Remove 4.8 single override from version bump script

omertuc:
It is no longer required because 4.8 is already part of the default versions JSON
as of this PR https://github.com/openshift/assisted-service/pull/1518


djzager:
This should be closed in favor of #90 but I don't have any power here.

PR #71, opened 2021-04-27
supporting clusters with the old 'metdata.json' files

osherdp:
Not really tested, but that should try to access the old 'metdata.json' files if 'metadata.json' files are not existing.


osherdp:
we'll try to rename the files instead

PR #88, opened 2021-05-13
MGMT-6062 add usage information to triage tickets

slaviered:


tools/add_triage_signature.py
```diff
@@ -262,6 +264,27 @@ class FailureDescription(Signature):
...
 
+    @staticmethod
+    def isOlm(name):
+        return name.lower() in ["cnv", "lso", "ocs"]
+
+    @staticmethod
+    def isGeneralFeature(cls, name):
+        return not cls.isOlm(name)
+
+    @staticmethod
```
osherdp, line 275: don't use static methods
ignore those warnings if the IDE doesn't like that :)

tools/add_triage_signature.py
```diff
@@ -273,7 +296,9 @@ def build_description(self, url, cluster_md):
                         "status": cluster_md['status'],
                         "status_info": cluster_md['status_info'],
                         "OCP_cluster_id": cluster_md['openshift_cluster_id'],
-                        "username": cluster_md['user_name']}
+                        "username": cluster_md['user_name'],
+                        "operators": self.build_feature_description(cluster_md, self.isOlm),
+                        "features": self.build_feature_description(cluster_md, self.isGeneralFeature)}
```
osherdp, line 301: let's do this simpler
create a method that returns two lists, ``operators`` and ``general_features`` based on the ``isOlm`` function alone

osherdp (changes requested):

tools/add_triage_signature.py
slaviered: It's actually simpler this way, because in the way you suggested I'll need a temp storage for those lists because the entries still need to be distinct. Don't you agree?

osherdp: Not really, as an example:
```
In [14]: all_features = set(["single-node", "additional-ntp", "CNV", "LSO"])

In [15]: operators = set(feature for feature in all_features if isOlm(feature))

In [16]: general_features = all_features - operators

In [17]: operators
Out[17]: {'CNV', 'LSO'}

In [18]: general_features
Out[18]: {'additional-ntp', 'single-node'}
```

osherdp (changes requested):

tools/add_triage_signature.py
```diff
@@ -262,7 +264,36 @@ class FailureDescription(Signature):
...
+        It's keys are features names where feature is any capacity
+        configured by the user (including Olm operators).
+        For sake of display, operators and other features are listed
+        separately
+        """
+        feature_usage_field = cluster.get('feature_usage')
+        if not feature_usage_field:
+            return {"operators": "n/a", "features": "n/a"}
+
+        def format_features(features):
```
osherdp, line 284: define outside

tools/add_triage_signature.py
```diff
@@ -262,7 +264,36 @@ class FailureDescription(Signature):
     def __init__(self, jira_client):
         super().__init__(jira_client, comment_identifying_string="")
 
+    @staticmethod
```
osherdp, line 267: no static methods porfavor

tools/add_triage_signature.py
```diff
@@ -262,7 +264,36 @@ class FailureDescription(Signature):
...
+        def format_features(features):
+            if len(features) > 0:
+                return ', '.join(features)
+            else:
+                return "None"
+
+        feature_usage = json.loads(feature_usage_field)
+        operators = [feature for feature in feature_usage.keys() if cls.is_olm_operator(feature)]
+        other_features = [feature for feature in feature_usage.keys() if not cls.is_olm_operator(feature)]
+        return {"operators": format_features(operators), "features": format_features(other_features)}
```
osherdp, line 293:
```suggestion
        return format_features(operators), format_features(other_features)
```

tools/add_triage_signature.py
```diff
@@ -262,7 +264,36 @@ class FailureDescription(Signature):
...
+
     def build_description(self, url, cluster_md):
+        feature_description = self.build_feature_description(cluster_md)
```
osherdp, line 296:
```suggestion
        operators, features = self.build_feature_description(cluster_md)
```

osherdp (changes requested):

osherdp (approved):

PR #96, opened 2021-06-20
Add signature to check for recently common CNI configuration error

omertuc:
A lot of recent production installation failures seem to be caused by an
error where kubelet complains about CNI plugin configuration.

This new signature looks for this repetitive error in the kubelet
journal of one of the control plane nodes, and if it finds it more
than threshold (1000) times, it creates the signature.

The journal logs are found in the installer-gather bootstrap log bundle,
under the control-plane masters-gather directory.

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -124,6 +124,11 @@ def get_host_log_file(triage_logs_tar, host_id, filename):
 
     return triage_logs_tar.get(f"{hostname}.tar.gz/logs_host_{host_id}/{filename}")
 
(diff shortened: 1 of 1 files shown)


osherdp:
@omertuc some PR description will be much appreciated

omertuc:
Updates description

osherdp (approved):

osherdp:
tested? can I push?

omertuc:
Yep

omertuc:
Might want to retroactively run this on previous tickets... not sure who's responsible for doing that

osherdp:
it should run it on anything which still has logs

PR #100, opened 2021-06-22
Run the CNI signature per host

omertuc:
The CNI signature used to run only on the first host. I've modified it
to look for the logs on all hosts rather than just the first one

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -8,6 +8,8 @@ import logging
 import netrc
 import os
 import re
+
+import itertools
 import sys
 import tempfile
 from collections import OrderedDict, defaultdict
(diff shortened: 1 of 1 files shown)


tools/add_triage_signature.py
```diff
@@ -503,6 +505,29 @@ class CNIConfigurationError(Signature):
     def __init__(self, jira_client):
         super().__init__(jira_client, comment_identifying_string="h1. No CNI configuration")
 
+    @staticmethod
+    def _get_host_remote_hosts_ip_addresses(host):
```
osherdp, line 509: so just ``_get_neighbors``? or ``_get_host_neighbors``

tools/add_triage_signature.py
```diff
@@ -503,6 +505,29 @@ class CNIConfigurationError(Signature):
...
+        """
+        Given a host object, return a list of all IP addresses of the remote hosts in this host's
+        connectivity stanza.
```
osherdp, line 512: A bit of confusing documentation to be honest

tools/add_triage_signature.py
```diff
@@ -503,6 +505,29 @@ class CNIConfigurationError(Signature):
...
+        """
+        try:
+            return [(remote_host['host_id'], remote_host['l3_connectivity'][0]['remote_ip_address'])
+                    for remote_host in
+                    json.loads(host["connectivity"])['remote_hosts']]
+        except Exception:
```
osherdp, line 518: in which case we might have this exception?

omertuc: KeyError mostly, malformed service output :man_shrugging:

osherdp (approved):

PR #104, opened 2021-06-22
Add matching error signatures to the 'impact function' field

ronniel1:


tools/add_triage_signature.py
```diff
@@ -168,6 +169,18 @@ def update_ticket(self, url, issue_key, should_update=False):
...
 
+    def _add_signature_label(self, key, labels):
+        issue_labels = self._jclient.issue(key).fields.__dict__[CF_FUNCTION_IMPACT]
+        if issue_labels is None:
+            issue_labels = []
+        is_changed = False
+        for l in labels:
+            if l not in issue_labels:
+                is_changed = True
+                issue_labels.append(l)
```
omertuc, line 180: The `for` loop can be replaced with:
```python
all_labels = set(issue_labels).union(labels)
if all_labels != issue_labels:
    self._update_fields(key, {CF_FUNCTION_IMPACT: all_labels})
```

omertuc:
lgtm other than nit

tools/add_triage_signature.py
ronniel1: yea, I actually thought about doing something like this, but desided it is less efficient and less clear
If you insist I will change

omertuc: I personally find for loops and flags much less clear, and clarity is more important than efficiency in code like this

Not insisting though, doesn't really matter

PR #106, opened 2021-06-23
NOISSUE: fixed close_by_signature.py

RazRegev:
Get transition id from the ticket's available transitions.
Fixed get_issues function.

### DIFF
--- a/tools/close_by_signature.py
+++ b/tools/close_by_signature.py
@@ -110,12 +110,11 @@ class IssueData:
         self.comment = comment
 
 
-def run_using_json(path, username, jira, issues, dry_run_stdout=None):
+def run_using_json(path, jira, issues, dry_run_stdout=None):
     logger.debug('Starting issue resolver task using filters json=%s', path)
     filters_json = read_filters_file(path)
     filters = get_filters_from_json(filters_json, jira)
     close_tickets_by_filters(
-        username=username,
         jira=jira,
         filters=filters,
         issues=issues,
@@ -198,14 +197,13 @@ def get_dry_run_stdout(args):
         return sys.stdout
 
 
-def close_tickets_by_filters(username, jira, filters, issues, dry_run_stdout):
+def close_tickets_by_filters(jira, filters, issues, dry_run_stdout):
     filtered_issues_generator = filter_and_generate_issues(
         jira=jira,
         filters=filters,
         issues=issues,
     )
     close_and_link_issues(
-        username=username,
         jira=jira,
         filtered_issues_generator=filtered_issues_generator,
         dry_run_stdout=dry_run_stdout,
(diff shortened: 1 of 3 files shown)


tools/close_by_signature.py
```diff
@@ -268,6 +278,59 @@ def link_issue_to_root_issue(jira, issue, root_issue, dry_run_stdout):
...
+    except (
+            requests.Timeout,
+            requests.RequestException,
+            requests.ConnectionError
+    ):
+        logger.exception('Failed to get available transitions for issue=%s',
+                         issue.key)
+
+
+TARGET_TRANSITION_NAME = 'Closed'
```
osherdp, line 318: maybe change to be a parameter in ``get_target_transition_id``

tools/close_by_signature.py
```diff
@@ -268,6 +278,59 @@ def link_issue_to_root_issue(jira, issue, root_issue, dry_run_stdout):
         )
 
 
+def close_issue(username, password, jira, issue):
+    available_transitions = get_available_transitions_for_issue(
+        username=username,
+        password=password,
+        issue=issue,
+    )
+    if not available_transitions:
```
osherdp, line 287: add a log about having no transition options

tools/create_triage_tickets.py
```diff
@@ -139,6 +139,7 @@ def main(arg):
     close_by_signature.run_using_json(
         path=args.filters_json,
         username=username,
+        password=password,
```
osherdp, line 142: also re-enable ticket closing:
https://github.com/openshift-assisted/assisted-installer-deployment/pull/98/files

osherdp (changes requested):

tools/close_by_signature.py
```diff
@@ -268,6 +278,65 @@ def link_issue_to_root_issue(jira, issue, root_issue, dry_run_stdout):
...
+
+
+API_VERSION = '2'
+REQUESTS_TIMEOUT = 5
+
+
+def get_available_transitions_for_issue(username, password, issue):
+    try:
+        resp = requests.get(
+            f'{JIRA_SERVER}/rest/api/{API_VERSION}/issue/{issue.key}/transitions',
```
ronniel1, line 313: instead of passing around the password, why not user 'jclient.transitions()'?

tools/close_by_signature.py
```diff
@@ -268,6 +265,28 @@ def link_issue_to_root_issue(jira, issue, root_issue, dry_run_stdout):
...
+        return
+
+    jira.transition_issue(issue, transition_id)
+
+
+def get_target_transition_id(jira, issue, target_name):
+    for transition in jira.transitions(issue):
+        assert 'name' in transition and 'id' in transition
+
+        if transition['name'] == target_name:
```
ronniel1, line 283: Why do we need to find the transiion id each time? Why not just add it as a constant here?

RazRegev: as I understood, transition ids can be unique per workflow, not necessarily all tickets share the same workflow:

https://community.atlassian.com/t5/Jira-Core-Server-questions/Is-there-is-a-list-of-all-available-transition-IDs/qaq-p/406492

tools/create_triage_tickets.py
```diff
@@ -162,6 +160,11 @@ def main(arg):
              '{signature_type: {root_issue: message}}',
         default='./triage_resolving_filters.json',
     )
+    parser.add_argument(
```
ronniel1, line 163: The argument name is misleading. because we are assigning only closed tickets
Just add it as a constant
.

PR #110, opened 2021-06-27
move CI tickets to AITRIAGE project on Jira

osherdp:
* Moving CI tickets creation to be against AITRIAGE project in oppose to MGMT
* Fixing ``jira_cmd.py`` code to look on triage tickets on AITRIAGE project
* Allow providing username and password via CLI on ``jira_cmd.py`` script

### DIFF
--- a/tools/jira_cmd.py
+++ b/tools/jira_cmd.py
@@ -53,11 +53,11 @@ def log_exception(msg):
         logger.error(msg)
 
 
-def jira_netrc_login(netrcFile):
+def jira_netrc_login(server, netrcFile):
     cred = netrc.netrc(os.path.expanduser(netrcFile))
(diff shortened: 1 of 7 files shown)


tools/jira_cmd.py
```diff
@@ -53,11 +53,11 @@ def log_exception(msg):
         logger.error(msg)
 
 
-def jira_netrc_login(netrcFile):
+def jira_netrc_login(server, netrcFile):
```
ronniel1, line 56: Why do you need this change?

ronniel1 (approved):

tools/jira_cmd.py
osherdp: my password contains a space
so netrc cannot contain it :(

PR #121, opened 2021-08-17
install skipper on build docker image

osherdp:


eranco74:
What for?

osherdp:
I think now we don't need it
I'll revert when we're finish with the CI setup

PR #126, opened 2021-08-18
fix all pep8 violations

osherdp:
Just going over the code to make it prettier (and make it pass the lint job).


YuviGold:
What's this `dco`?

osherdp:
> What's this `dco`?

won't be on next PRs
It's one of the plugins that I removed

YuviGold:
> > What's this `dco`?
> 
> won't be on next PRs
> It's one of the plugins that I removed

Okay. In that case `tide` still blocking just override

PR #128, opened 2021-08-28
Overcome testgrid bug where grids with green 'Overall' are marked as FAILING

omertuc:
See comments in diff for more information

### DIFF
--- a/tools/create_testgrid_tickets.py
+++ b/tools/create_testgrid_tickets.py
@@ -65,9 +65,13 @@ def get_last_failure_instance(test_id):
     return res.json()['changelists'][0]
 
 
-def get_last_pass(test_data):
+def get_overall_tests(test_data):
     return [t for t in test_data['tests']
-            if t['display_name'] == 'Overall'][0]['pass_timestamp']
+            if t['display_name'] == 'Overall']
+
+
+def get_last_pass(test_data):
+    return get_overall_tests(test_data)[0]['pass_timestamp']
 
 
 def format_labels(failure_data):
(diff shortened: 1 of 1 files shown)


tools/create_testgrid_tickets.py
```diff
@@ -111,11 +115,24 @@ def create_jira_ticket(jclient, existing_tickets, test_id, test_data):
...
 
+def is_test_failing(test_data):
+    failing = test_data['overall_status'] == "FAILING"
+
+    # testgrid has a bug where it marks successful tests as "FAILING", just because they
+    # have failed test cases. But sometimes test cases are allowed to fail while the run itself is
+    # "passing", but testgrid doesn't take that into account - so it marks a grid as "FAILING" even though
+    # its "Overall" row is green. When the "Overall" row is green but the test is marked as failing, the 
+    # overall tests don't get included in the `tests` list, so we use that to detect this scenario.
+    actually_failing = len(get_overall_tests(test_data)) > 0
```
osherdp, line 126: So let's just make this comparison
Having both failing and actual_failing is just super confusing

osherdp (changes requested):

osherdp (approved):

tools/create_testgrid_tickets.py
omertuc: Fixed. I think this is super confusing because the bug is super confusing, my 2 variables are not the problem :upside_down_face:

osherdp:
/ok-to-test
(you're not on reviewers list?)

omertuc:
nope

tools/create_testgrid_tickets.py
osherdp: In case of duality due to some outside system, there's no need to add additional duality inside, that's all I'm saying 🙂

osherdp (approved):

PR #130, opened 2021-08-31
Update approvers list to include Omer and Toni

osherdp:
/cc @omertuc @celebdor


osherdp:
/hold
(we'll first add them to the org)

PR #132, opened 2021-09-01
Use skopeo to verify assisted-installer.yaml snapshot

YuviGold:
/cc @osherdp 
/cc @oshercc

### DIFF
--- a/tools/check_ai_images.py
+++ b/tools/check_ai_images.py
@@ -1,9 +1,10 @@
-import os
-import yaml
-import time
 import argparse
 import logging
+import os
 import subprocess
+
+import yaml
+from retry import retry
 
 parser = argparse.ArgumentParser()
 parser.add_argument("--deployment", help="deployment yaml file to update", type=str,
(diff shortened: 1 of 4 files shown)


tools/check_ai_images.py
```diff
@@ -14,41 +15,30 @@
...
-            return True
-        logger.warning(f"Failed to get image at {retry} retry")
-        time.sleep(10*60)
-
-    logger.error(f"Failed to get image at {retry} retry")
-    return not get_image_result
+@retry(exceptions=subprocess.SubprocessError, tries=3, delay=10)
+def does_image_exist(pull_spec: str):
+    cmd = f"skopeo inspect --config docker://{pull_spec}"
+    return subprocess.check_call(cmd, stdout=subprocess.PIPE, shell=True) == 0
```
omertuc, line 24: ```check_call() and check_output() will raise CalledProcessError if the called process returns a non-zero return code.```

omertuc: >Note Do not use stdout=PIPE or stderr=PIPE with this function. The child process will block if it generates enough output to a pipe to fill up the OS pipe buffer as the pipes are not being read from.

omertuc: There's no reason to compare `== 0` because check_call always returns 0

omertuc: `subprocess.call` matches the behavior you're looking for

osherdp (approved):

tools/check_ai_images.py
```diff
@@ -14,41 +15,27 @@
...
-        if not get_image_result:
-            return True
-        logger.warning(f"Failed to get image at {retry} retry")
-        time.sleep(10*60)
-
-    logger.error(f"Failed to get image at {retry} retry")
-    return not get_image_result
+@retry(exceptions=subprocess.SubprocessError, tries=3, delay=10)
+def does_image_exist(pull_spec: str) -> bool:
+    return subprocess.call(f"skopeo inspect --config docker://{pull_spec}", stdout=subprocess.DEVNULL, shell=True) == 0
```
omertuc, line 23: It's better not to use `shell=True` with variable input (`pull_spec`), because if it contains shell characters such as `\&{}` it'll cause trouble. Better to just `["skopeo", "inspect", "--config", "docker://{pull_spec}"]`

osherdp: yeah let's replace it
@omertuc you wanna do it real quick?

omertuc: https://github.com/openshift-assisted/assisted-installer-deployment/pull/133

PR #136, opened 2021-09-05
Enable providing arguments via environment variables

osherdp:
@YuviGold


YuviGold:
I think it will fail the Jenknisfile. You need to change it as well https://github.com/openshift-assisted/assisted-installer-deployment/blob/master/Jenkinsfile.default_release_update

osherdp:
@YuviGold how should it fail this job that you mentioned?

YuviGold:
> @YuviGold how should it fail this job that you mentioned?

You are right, I thought it replaced the old API. :)

YuviGold (approved):

PR #147, opened 2021-10-17
Adding a new image to distinguish build and 'tools' purposes

osherdp:
/cc @eliorerz


Dockerfile.assisted-installer-deployment
```diff
@@ -0,0 +1,29 @@
+FROM quay.io/centos/centos:stream8
+
+RUN dnf update -y && dnf install -y \
+    jq \
+    gcc \
+    git \
+    make \
+    skopeo \
+    python38 \
+    python38-pip \
```
eliorerz, line 10: Why python 3.8 and not 3.9?

osherdp: you're right :)

eliorerz (approved):
LGTM

PR #158, opened 2021-11-18
Refactor some of the release scripts

osherdp:
A couple of things I needed to have for a successful release process:
* Adding bugzilla library dependency that was just missing
* Starting to ditch the build image, that's missing actual code. We will need to remove it from release repo as well
* Mounting ``assisted-installer.yaml`` and ``netrc`` to allow using them from within skipper
* Some minor changes like using f-strings
* Skipping update of tickets from project ``OCPBUGSM``, which is non-writable

/cc @eliorerz 
/hold


eliorerz (approved):
LGTM

PR #162, opened 2021-12-01
[WIP] Change JIRA authentication from Basic to PAT

mkowalski:
/cc @omertuc 
/cc @osherdp


osherdp:
/close
I guess we can close as #163 is in

PR #166, opened 2021-12-06
Allow specifying more domains for automatic AITRIAGE closure

mkowalski:


osherdp (approved):

osherdp:
/lgtm
/approve
/hold
all in all, sounds good to me
maybe just update the PR description to explain the reasoning behind it

osherdp (approved):

tools/add_triage_signature.py
```diff
@@ -32,6 +32,7 @@
 CF_DOMAIN = "12319045"
 CF_CLUSTER_ID = "12316349"
 CF_FUNCTION_IMPACT = "12317358"
+CF_IGNORED_DOMAINS = ["redhat.com", "juniper.net"]
```
omertuc, line 35: should be { } set not [ ] list

omertuc: (nit, of-course. it's insanely negligible performance-wise, it's just makes the intent more clear)

osherdp (approved):

mkowalski:
We are getting >10 tickets per day that should be handled by the change in this PR. I'm going to forcefully unhold as per discussion, we can't easily make the repo private without breaking the surrounding infrastructure.

Not ideal, but it is what it is

/unhold

PR #167, opened 2021-12-07
Update image verification script to use new image repository

osherdp:
Make docker registry and repository part of specification of images location, and update code to use the new edge-infrastructure quay.io organization. Images in the new organization should follow a different pattern ``latest-<hash>`` so making the code taking that into account.
For now, only ``assisted-service`` and ``assisted-image-service`` have constant image mirroring, so until handling other images we'll just update what we have.

### DIFF
--- a/assisted-installer.yaml
+++ b/assisted-installer.yaml
@@ -1,24 +1,24 @@
 openshift/assisted-installer:
-  revision: c81363a7fada8d23cfbac86ecc0f97835f1fa043
+  revision: 2b5e964a1a39debf53516ebfc6e5c7572d4f9987
   images:
-  - assisted-installer-controller
-  - assisted-installer
+  - quay.io/edge-infrastructure/assisted-installer-controller
+  - quay.io/edge-infrastructure/assisted-installer
 openshift/assisted-service:
-  revision: b9ab5d3c77580ab93c29bb43cb78cc6b02d54e88
+  revision: 364b9d71b11250ae536787ef354c2d7cb516bb7c
   images:
-  - assisted-service
+  - quay.io/edge-infrastructure/assisted-service
 openshift-assisted/assisted-ui:
-  revision: d60808caecfdb4f6c480dd8e5b07213ea4f210dd
+  revision: b5281fe43c01b2f8adee5483b3bb438bf487cb91
   images:
-  - ocp-metal-ui
+  - quay.io/ocpmetal/ocp-metal-ui
 openshift/assisted-installer-agent:
-  revision: 1b51ecd2786c0f56ceb8f2051fe0af4dae1f7173
+  revision: 55d9252ac2bb120f19a7779b8f5a23ef4c1fe048
   images:
-  - assisted-installer-agent
+  - quay.io/edge-infrastructure/assisted-installer-agent
 openshift/assisted-image-service:
-  revision: a6f4e9968f11ac730523a38fe17bac0661a885ed
+  revision: 725ace298b84e71cfa4be6be364f7483692cc1e0
   images:
-  - assisted-image-service
+  - quay.io/edge-infrastructure/assisted-image-service
 
 # Downstream images:
 # - assisted-installer:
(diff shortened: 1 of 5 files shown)


osherdp:
/hold
(to let people review code)

lranjbar:
What do you mean only assisted-service and assisted-image-service have only constant image mirroring? They all should.

assisted-installer.yaml
```diff
@@ -1,24 +1,24 @@
 openshift/assisted-installer:
   revision: c81363a7fada8d23cfbac86ecc0f97835f1fa043
   images:
-  - assisted-installer-controller
-  - assisted-installer
+  - quay.io/ocpmetal/assisted-installer-controller
+  - quay.io/ocpmetal/assisted-installer
 openshift/assisted-service:
```
lranjbar, line 6: This should be edge-infrastructure..

assisted-installer.yaml
```diff
@@ -1,24 +1,24 @@
...
   revision: b9ab5d3c77580ab93c29bb43cb78cc6b02d54e88
   images:
-  - assisted-service
+  - quay.io/edge-infrastructure/assisted-service
 openshift-assisted/assisted-ui:
   revision: d60808caecfdb4f6c480dd8e5b07213ea4f210dd
   images:
-  - ocp-metal-ui
+  - quay.io/ocpmetal/ocp-metal-ui
```
lranjbar, line 13: This is the only one I know hasn't moved.

assisted-installer.yaml
```diff
@@ -1,24 +1,24 @@
...
 openshift/assisted-installer-agent:
   revision: 1b51ecd2786c0f56ceb8f2051fe0af4dae1f7173
   images:
-  - assisted-installer-agent
+  - quay.io/ocpmetal/assisted-installer-agent
```
lranjbar, line 17: This one is in edge-infrastructure as well.

lranjbar (changes requested):
I think this is more of the case that we haven't pushed changes to these repos in a long time. This is why there isn't a GIT_SHA tagged image associated with it. Do you want to copy the latest one over and call it done?

assisted-installer.yaml
osherdp: right
fixed!

assisted-installer.yaml
osherdp: yeah, maybe it also not that interesting
to the best of my knowledge, the UI team has a very different flow for releasing and deploying their image (rather as JS package), plus this repo doesn't use prow

assisted-installer.yaml
osherdp: indeed ``assisted-installer`` can be changed, ``assisted-installer-controller`` not yet
once we'll trigger image-mirroring we can change it also

osherdp: fixed

tools/check_ai_images.py
```diff
@@ -11,7 +11,7 @@
 logger = logging.getLogger(__name__)
 logging.getLogger("__main__").setLevel(logging.INFO)
 
-DEFAULT_REGISTRY = "quay.io/ocpmetal"
+DEFAULT_REGISTRY = "quay.io/edge-infrastructure"
```
carbonin, line 14: Is this used anywhere now?

nshidlin:
@osherdp will https://github.com/openshift-assisted/assisted-installer-deployment/blob/ece8d840b97b295720ad3f91447306a7bf6abf8b/tools/assisted_installer_stable_promotion.py#L13 come in separate PR?

osherdp:
> @osherdp will
> 
> https://github.com/openshift-assisted/assisted-installer-deployment/blob/ece8d840b97b295720ad3f91447306a7bf6abf8b/tools/assisted_installer_stable_promotion.py#L13
> 
> come in separate PR?

no, it would have broken it. good catch

tools/check_ai_images.py
osherdp: yes, by QE
I'll work together with @nshidlin to make the transition. I think it doesn't make sense to block on it because it seems co-depending with this change

assisted-installer.yaml
osherdp: okay, it should be handled by UI team very soon. We'll update this location once they finish doing that

PR #168, opened 2021-12-15
Change UI image location to edge-infrastructure org

osherdp:
/cc @nshidlin


osherdp:
@jkilzi

assisted-installer.yaml
```diff
@@ -10,7 +10,7 @@ openshift/assisted-service:
 openshift-assisted/assisted-ui:
   revision: b5281fe43c01b2f8adee5483b3bb438bf487cb91
```
nshidlin, line 11: does the revision need to be bumped? I ask as you changed this in your last PR

osherdp:
/hold
I'll check if it needs to bump the revision
although we can also just run the jenkins job

osherdp:
@nshidlin updated the revisions as well

PR #170, opened 2021-12-23
MGMT-8400: remove jira basic-auth method

osherdp:
Removing old authentication method, as our Jira instance no longer supports it.


eliorerz (approved):
LGTM

PR #171, opened 2021-12-23
copy OCP bump script from assisted-service

osherdp:
Move OCP bump script, so we can have all management Python scripts in one place.
Ultimately, it will also help moving this job from Jenkins to prow.


eliorerz (approved):
LGTM

PR #173, opened 2022-01-05
Add script for producing CI status reports

osherdp:
Plot a bar-chart of our prow CI jobs states to slack.
jobs

/cc @eliorerz @omertuc 
/hold


tools/ci_status_report.py
```diff
@@ -0,0 +1,180 @@
...
+import dataclasses
+import itertools
+
+import requests
+import matplotlib.pyplot as plt
+
+
+PROW_JOBS_STATUSES = "https://prow.ci.openshift.org/prowjobs.js?omit=annotations,labels,decoration_config,pod_spec"
+SLACK_FILES_UPLOAD = "https://slack.com/api/files.upload"
+TRACKED_REPOSITORIES = list(itertools.chain(
```
osherdp, line 16: probably we should add some more

tools/ci_status_report.py
```diff
@@ -0,0 +1,180 @@
...
+            organization=organization,
+            repository=repository,
+            base_ref=base_ref,
+            state=State(document["status"]["state"]),
+            url=document["status"]["url"],
+            start_time=document["status"]["startTime"],
+            completion_time=document["status"].get("completionTime"),
+        )
+
+        if (job.organization, job.repository) in TRACKED_REPOSITORIES:
```
omertuc, line 101: Isn't this condition redundant?

tools/ci_status_report.py
```diff
@@ -0,0 +1,180 @@
...
+    organization: str
+    repository: str
+    base_ref: str
+    state: State
+    url: str
+    start_time: datetime.datetime
+    completion_time: datetime.datetime
+
+
+def filter_jobs(response):
```
omertuc, line 71: `items = response.json().get("items", {})`

`spec = document.get("spec", {})` 

To be more robust in the face of changed format

Also in the `job = Job` assignment, used `get` to avoid crashing on missing info

omertuc: (unless you want to crash on missing info which is also reasonable)

tools/ci_status_report.py
```diff
@@ -0,0 +1,180 @@
...
+            del status_counters[job]
+
+    return status_counters
+
+
+def draw_figure(status_counters):
+    _, ax = plt.subplots(figsize=(15,15))
+
+    for state, color in STATE_TO_COLOR.items():
+        ax.bar(
```
osherdp, line 125: have a bug here
the values are probably overlapping each other, instead of stacking

tools/ci_status_report.py
osherdp: no, not at all
for some reason, I couldn't query this API to get just jobs for specific repos/organizations. So it basically brings everything and we filter our stuff from it

omertuc: I should have emphasized - you already check this condition in 

```python
        for ref in all_refs:
            organization = ref["org"]
            repository = ref["repo"]
            base_ref = ref["base_ref"]
            if (organization, repository) in TRACKED_REPOSITORIES:
                break

        else:
            continue
```

osherdp: oh yes, you're right
I'll remove it

tools/ci_status_report.py
osherdp: I'll do that for the out scope, as it's indeed might be missing from the schema, and have everything around ``Job`` assignment with regular access, so we can crash on missing info

tools/ci_status_report.py
osherdp: fixed

osherdp:
@omertuc @eliorerz I made some changes to have it horizontal
when you have the time please take a look

eliorerz (approved):
LGTM

PR #175, opened 2022-01-08
NO-ISSUE: remove reports dir creation

osherdp:
Remove creation and usage of ``reports`` dir, as it's not really in use.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,8 +17,7 @@ autopep8:
 	autopep8 --recursive --in-place .
 
 pylint:
-	mkdir -p reports
-	PYLINTHOME=reports/ pylint release/
+	pylint release/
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -17,8 +17,7 @@ autopep8:
 	autopep8 --recursive --in-place .
 
 pylint:
-	mkdir -p reports
-	PYLINTHOME=reports/ pylint release/
+	pylint release/
```
eliorerz, line 20: Not related to this PR, But why using pylint and not flake8 as we use in test-infra?

osherdp: We do both

eliorerz (approved):
LGTM

PR #177, opened 2022-01-11
Added bypass-iso-download flag to bump_ocp_releases

danielerez:
To ensure dry-run mode invokes the entire flow except actually push a PR, added a new flag to bypass ISO download.
I.e. use --bypass-iso-download to avoid downloading the RHCOS isos (used for fetching build version).


osherdp (approved):
Perfect!

PR #179, opened 2022-01-12
remove CI tickets creation code

osherdp:
It's no longer used, as we have notifications and reports for it.
/cc @mkowalski @eliorerz 
/hold
(waiting for openshift/assisted-test-infra#1369 to get in)


eliorerz (approved):
LGTM

PR #181, opened 2022-01-16
Remove usage of yamlordereddictloader as dicts already ordered

osherdp:
As of Python 3.7, all dictionaries are already ordered by time of insertion, so this dependency is not needed anymore.
/cc @eliorerz


eliorerz (approved):
LGTM

PR #184, opened 2022-01-19
bump_ocp_releases: added generate-bundle

danielerez:
Added missing 'make generate-bundle'

### DIFF
--- a/Dockerfile.assisted-installer-deployment
+++ b/Dockerfile.assisted-installer-deployment
@@ -1,8 +1,13 @@
 FROM quay.io/centos/centos:stream8
 
+ENV GOPATH=/root/go
+ENV GOROOT=/usr/lib/golang
+ENV PATH=$PATH:$GOROOT/bin:$GOPATH/bin
+
 RUN dnf update -y && dnf install -y \
     jq \
     gcc \
+    golang \
     git \
     make \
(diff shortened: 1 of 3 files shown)


osherdp (approved):

Dockerfile.assisted-installer-deployment
```diff
@@ -30,3 +26,8 @@ WORKDIR /home/assisted-installer-deployment
 COPY . .
 
 RUN pip install .
+
+# required for generating configuration/bundle in assisted-service repo
+FROM registry.ci.openshift.org/openshift/release:golang-1.16
+COPY ./setup_env.sh ./
+RUN ./setup_env.sh assisted_service
```
osherdp, line 33: this is probably problematic, as it's probably changes the current directory. no?

osherdp: why we need to be based on ``registry.ci.openshift.org/openshift/release:golang-1.16``?

osherdp (changes requested):

Dockerfile.assisted-installer-deployment
danielerez: no mandatory, added to dnf install instead

osherdp (approved):

PR #186, opened 2022-01-30
couple of fixes to the release_tickets scripts

osherdp:
Fixing the following bugs:
* We changed the schema of jira versions to ``AI <version>``, so changing it accordingly in the code
* We mistakenly set fixed version even for fixes that are already included in previous patch-versions
* Code should refer to location of netrc file in the container, and not of the one in the host


eliorerz (approved):
LGTM

PR #187, opened 2022-02-02
MGMT-8784: bump versions - skip x.y.z

danielerez:
Avoid updating x.y.z versions as only x.y should updated to latest.

### DIFF
--- a/tools/bump_ocp_releases.py
+++ b/tools/bump_ocp_releases.py
@@ -62,6 +62,7 @@ ASSISTED_SERVICE_MASTER_DEFAULT_RELEASE_IMAGES_JSON_URL = \
 OCP_REPLACE_CONTEXT = ['"{version}"', "ocp-release:{version}"]
 
 SKIPPED_MAJOR_RELEASE = ["4.6"]
+MAJOR_MINOR_VERSION_REGEX = re.compile(r"^([1-9]\d*|0)(\.(([1-9]\d*)|0))$")
 
 CPU_ARCHITECTURE_AMD64 = "amd64"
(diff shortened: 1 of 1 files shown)


tools/bump_ocp_releases.py
```diff
@@ -62,6 +62,7 @@
 OCP_REPLACE_CONTEXT = ['"{version}"', "ocp-release:{version}"]
 
 SKIPPED_MAJOR_RELEASE = ["4.6"]
+MAJOR_MINOR_VERSION_REGEX = re.compile("^([1-9]\d*|0)(\.(([1-9]\d*)|0))$")
```
osherdp, line 65: lint probably wants
```suggestion
MAJOR_MINOR_VERSION_REGEX = re.compile(r"^([1-9]\d*|0)(\.(([1-9]\d*)|0))$")
```

osherdp: btw, doesn't it work too?
```suggestion
MAJOR_MINOR_VERSION_REGEX = re.compile("^[1-9][0-9]*\.[1-9][0-9]*$")
```

osherdp (changes requested):

tools/bump_ocp_releases.py
danielerez: > lint probably wants

probably, changed.

danielerez: > btw, doesn't it work too?

it won't match 5.0 for example

osherdp: right

osherdp (approved):
/hold
unhold when the job passes

PR #189, opened 2022-02-08
Make 404 & isoinfo errors more indicative

osherdp:
We had two problems:
* when we have 404 errors (for example) we do not raise
* when we fail on ``isoinfo``, we have no stderr output

/cc @danielerez 
/hold


osherdp:
/override ci/prow/bump-ocp-dry-run
(known failure. we'll have to wait for ARTists to fix rc.1 artifacts)

PR #197, opened 2022-03-08
get vcs ref from latest image

rccrdpccl:
This change aims to eliminate possibility of race conditions on pushing images to quay.io and calculating the latest vcs ref available.

Instead of fetching the latest ref from HEAD through git and then checking if the image is ready, we will get the latest tag matching `latest-<git hash>`.
Unfortunately we have some issue with image tagging:
* not all docker images mantain `vcs-ref` label. And even if they did, there are other issues that might break the job if we go that direction (see below)
* not all `latest-<git hash>` tags are generated. Sometimes the process glitches and as result the tag is never generated
* latest tag is not always pointing (having the vcs-ref label) at the latest ref available (mirroring to latest fails - although somewhat rarely). This is not a big issue per se, as worst case scenario we will just snapshot a previous version

NOTE: this script is used in other parts of the CI:
https://github.com/openshift/release/blob/master/ci-operator/config/openshift-assisted/assisted-installer-deployment/openshift-assisted-assisted-installer-deployment-master.yaml#L77
https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift-assisted/assisted-installer-deployment/openshift-assisted-assisted-installer-deployment-master-presubmits.yaml#L216

It looks like those steps might also benefit from the same change, however if it's not so please let me know so we can change this PR accordingly.

### DIFF
--- a/tools/skopeo.py
+++ b/tools/skopeo.py
@@ -1,0 +1,18 @@
+import json
+import subprocess
+import re
+from retry import retry
+
+
+class Skopeo:
+    @retry(exceptions=subprocess.SubprocessError, tries=3, delay=10)
+    def get_image_tags_by_pattern(self, repository: str, pattern: str) -> list[str]:
+        """Returns prefixed tags ordered by most recent to least recent"""
+        docker_uri = f"docker://{repository}"
+        out = subprocess.check_output(["skopeo", "list-tags", docker_uri])
+        json_output = out.decode("utf-8")
+        response = json.loads(json_output)
+        if "Tags" not in response:
+            return []
+        # tags are returned by skopeo from oldest to newest. Let's reverse the order
+        return [tag for tag in response["Tags"] if re.search(pattern, tag) is not None][::-1]
--- a/tools/update_assisted_installer_yaml.py
+++ b/tools/update_assisted_installer_yaml.py
@@ -1,14 +1,42 @@
 import argparse
-import subprocess
 import yaml
 import os
 import update_hash
+import skopeo
 
 parser = argparse.ArgumentParser()
 parser.add_argument("--deployment", help="deployment yaml file to update", type=str,
                     default=os.path.join(os.path.dirname(__file__), "../assisted-installer.yaml"))
 parser.add_argument('--full', help='update all hashes to master', action='store_true')
 args = parser.parse_args()
+
+
+def find_first_common_element(lists: list[str]) -> str:
+    if len(lists) < 1:
+        raise ValueError("No lists provided. Cannot find common element")
+
+    for el in lists[0]:
+        if all(el in list for list in lists[1:]):
+            return el
+
+    return None
+
+
+def get_ref_by_docker_image(images):
+    tag_pattern = "^latest-[a-f0-9]{40}$"
+    tag_prefix = "latest-"
+    skopeo_client = skopeo.Skopeo()
+    image_tags = [skopeo_client.get_image_tags_by_pattern(image, tag_pattern)
+                  for image in images]
+
+    if len(image_tags) == 0:
+        raise IndexError(f"Could not find any tags matching pattern `{tag_pattern}` for images `{images}`")
+
+    tag = find_first_common_element(image_tags)
+    if tag is None:
+        raise ValueError(f"Couild not find common tag for images {images}")
+
+    return tag.removeprefix(tag_prefix)
 
 
 def main():
@@ -19,13 +47,11 @@ def main():
 
     for rep in deployment:
         if full_release:
-            hash = subprocess.check_output(f"git ls-remote https://github.com/{rep}.git HEAD | cut -f 1", shell=True)[:-1]
-            hash = hash.decode("utf-8")
+            hash = get_ref_by_docker_image(deployment[rep]['images'])
         else:
             hash = os.environ.get(rep, None)
-            if not hash:
-                continue
-
+        if not hash:
+            continue
         update_hash.update_hash(deployment_yaml=args.deployment, repo=rep, hash=hash)
 
 


rccrdpccl:
cc @michaellevy101 @osherdp

Dockerfile.assisted-installer-deployment
```diff
@@ -15,6 +15,7 @@ RUN dnf update -y && dnf install -y \
     python39-pip \
     python39-devel \
     genisoimage \
+    skopeo \
```
osherdp, line 18: seems like we already install it on line 13

tools/update_assisted_installer_yaml.py
```diff
@@ -11,6 +12,26 @@
 args = parser.parse_args()
 
 
+def get_ref_by_docker_image(images):
+    images_attrs = []
+    for img in images:
+        docker_uri = "docker://" + img + ":latest"
```
osherdp, line 18:
```suggestion
        docker_uri = f"docker://{img}:latest"
```

tools/update_assisted_installer_yaml.py
```diff
@@ -11,6 +12,26 @@
...
+        output = subprocess.run(["skopeo", "inspect", docker_uri], capture_output=True)
+        json_out = output.stdout.decode("utf-8")
```
osherdp, line 20:
```suggestion
        output = subprocess.check_output(["skopeo", "inspect", docker_uri])
        json_out = output.decode("utf-8")
```

tools/update_assisted_installer_yaml.py
```diff
@@ -11,6 +12,26 @@
...
+        image = json.loads(json_out)
+        created_at = image['Created']
+        vcs_ref = image['Labels']['vcs-ref']
+        images_attrs.append({"ref": vcs_ref, "created_at": created_at})
+
+    # if all references are the same, just return it
+    if all(item["ref"] == images_attrs[0]["ref"] for item in images_attrs):
+        return images_attrs[0]["ref"]
+
+    # if they differ, let's return the oldest one, as it is likely to be common
```
osherdp, line 30: removing it will yield the same results, no?
```suggestion
```

osherdp (changes requested):

tools/update_assisted_installer_yaml.py
```diff
@@ -11,22 +12,56 @@
...
+def in_all_lists(el: str, lists: list[str]) -> bool:
+    for list in lists:
+        if el not in list:
+            return False
+    return True
+
+
+def find_first_common_element(lists: list[str]) -> str:
+    if len(lists) < 1:
+        raise IndexError("No lists provided. Cannot find common element")
```
osherdp, line 24: Seems to be more relevant in this case:
```suggestion
        raise ValueError("No lists provided. Cannot find common element")
```

tools/update_assisted_installer_yaml.py
```diff
@@ -11,22 +12,56 @@
...
+    for el in lists[0]:
+        if in_all_lists(el, lists[1:]):
+            return el
+    raise RuntimeError("Could not find common element")
```
osherdp, line 28: it seems more right to return ``None`` in this case:
```suggestion
    return None
```

tools/update_assisted_installer_yaml.py
```diff
@@ -11,22 +12,56 @@
...
+
+
+def get_ref_by_docker_image(images):
+    tag_pattern = "^latest-[a-f0-9]{40}$"
+    tag_prefix = "latest-"
+    sk = skopeo.Skopeo()
```
osherdp, line 34: can we change all names here to not include abbrev?
e.g. ``sk->skeopeo_client``, ``img->image``
it will make code more English-like

tools/update_assisted_installer_yaml.py
```diff
@@ -11,22 +12,56 @@
...
+    image_tags = []
+    for img in images:
+        image_tags.append(sk.get_image_tags_by_pattern(img, tag_pattern))
```
osherdp, line 37:
```suggestion
    image_tags = [sk.get_image_tags_by_pattern(img, tag_pattern)
                  for img in images]
```

tools/update_assisted_installer_yaml.py
```diff
@@ -11,22 +12,56 @@
...
+            else:
+                hash = os.environ.get(rep, None)
             if not hash:
                 continue
-
-        update_hash.update_hash(deployment_yaml=args.deployment, repo=rep, hash=hash)
+            update_hash.update_hash(deployment_yaml=args.deployment, repo=rep, hash=hash)
+        except LookupError as e:
+            print(e, file=sys.stderr)
+            sys.exit(1)
```
osherdp, line 64: seems to be equivalent to just not ``try-excepting`` and letting it fail for this exception

tools/update_assisted_installer_yaml.py
```diff
@@ -11,22 +12,56 @@
...
+
+
+def get_ref_by_docker_image(images):
+    tag_pattern = "^latest-[a-f0-9]{40}$"
+    tag_prefix = "latest-"
+    sk = skopeo.Skopeo()
+    image_tags = []
+    for img in images:
+        image_tags.append(sk.get_image_tags_by_pattern(img, tag_pattern))
+    if len(image_tags) < 1:
```
osherdp, line 38: I think it's more explicit to do:
```suggestion
    if len(image_tags) == 0:
```

tools/update_assisted_installer_yaml.py
```diff
@@ -11,22 +12,56 @@
...
+        raise IndexError(f"Could not find any tags matching pattern `{tag_pattern}` for images `{images}`")
+    if len(image_tags) == 1:
+        tag = image_tags[0][0]
```
osherdp, line 41: redundant, no?
I means, seems like removal of those lines will yield the same results
```suggestion
```

tools/skopeo.py
```diff
@@ -0,0 +1,18 @@
+import json
+import subprocess
+import re
+from retry import retry
+
+
+class Skopeo:
```
osherdp, line 7: Great idea to have this kind of abstraction :+1:

osherdp (changes requested):

tools/update_assisted_installer_yaml.py
```diff
@@ -11,22 +12,56 @@
...
+        if el not in list:
+            return False
+    return True
+
+
+def find_first_common_element(lists: list[str]) -> str:
+    if len(lists) < 1:
+        raise IndexError("No lists provided. Cannot find common element")
+    for el in lists[0]:
+        if in_all_lists(el, lists[1:]):
```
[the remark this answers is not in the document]
rccrdpccl: Thanks for the great feedback :wink:

tools/update_assisted_installer_yaml.py
```diff
@@ -11,6 +11,38 @@
 args = parser.parse_args()
 
 
+def in_all_lists(el: str, lists: list[str]) -> bool:
+    for list in lists:
+        if el not in list:
+            return False
+    return True
```
osherdp, line 18: this seems redundant now:
```suggestion
```

tools/update_assisted_installer_yaml.py
```diff
@@ -11,6 +11,38 @@
...
+    skopeo_client = skopeo.Skopeo()
+    image_tags = []
+    image_tags = [skopeo_client.get_image_tags_by_pattern(image, tag_pattern)
+                  for image in images]
+
+    if len(image_tags) == 0:
+        raise IndexError(f"Could not find any tags matching pattern `{tag_pattern}` for images `{images}`")
+    tag = find_first_common_element(image_tags)
+    if tag is None:
+        raise ValueError(f"Couild not find common tag for images {images}")
```
osherdp, line 42: nit: append blank line

tools/update_assisted_installer_yaml.py
```diff
@@ -11,6 +11,38 @@
...
+def get_ref_by_docker_image(images):
+    tag_pattern = "^latest-[a-f0-9]{40}$"
+    tag_prefix = "latest-"
+    skopeo_client = skopeo.Skopeo()
+    image_tags = []
+    image_tags = [skopeo_client.get_image_tags_by_pattern(image, tag_pattern)
+                  for image in images]
+
+    if len(image_tags) == 0:
+        raise IndexError(f"Could not find any tags matching pattern `{tag_pattern}` for images `{images}`")
```
osherdp, line 39: nit: append blank line

tools/update_assisted_installer_yaml.py
```diff
@@ -11,6 +11,38 @@
...
+def in_all_lists(el: str, lists: list[str]) -> bool:
+    for list in lists:
+        if el not in list:
+            return False
+    return True
+
+
+def find_first_common_element(lists: list[str]) -> str:
+    if len(lists) < 1:
+        raise ValueError("No lists provided. Cannot find common element")
```
osherdp, line 23: nit: append blank line

tools/update_assisted_installer_yaml.py
```diff
@@ -11,6 +11,38 @@
...
+    for el in lists[0]:
+        if all(el in list for list in lists[1:]):
+            return el
```
osherdp, line 26: nit: append blank line

tools/update_assisted_installer_yaml.py
```diff
@@ -11,6 +11,38 @@
...
+    return None
+
+
+def get_ref_by_docker_image(images):
+    tag_pattern = "^latest-[a-f0-9]{40}$"
+    tag_prefix = "latest-"
+    skopeo_client = skopeo.Skopeo()
+    image_tags = []
```
osherdp, line 34: you can delete it:
```suggestion
```

osherdp (changes requested):
just some small edits and we're ready

osherdp:
/hold
(just checking dry-run execution)

osherdp:
/unhold
yep, looks magnificent 
thank you @rccrdpccl!

PR #198, opened 2022-03-25
set default value to status.url when not present for prow jobs

rccrdpccl:
[Example failure](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-assisted-assisted-installer-deployment-master-ci-status-report/1507251112380469248)
Not sure how this did not happen before, we can see an example response for the request to "https://prow.ci.openshift.org/prowjobs.js?omit=annotations,labels,decoration_config,pod_spec" in the [tools/fixtures/prow_job_statuses.json](https://github.com/openshift-assisted/assisted-installer-deployment/compare/master...rccrdpccl:default-when-url-not-present?expand=1#diff-38a1e52ae19eff705979bba4b10d5692c1ce7f82a7ff0395e20a69d0e36e06ca) file - on which I've based the change on.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,9 @@ lint: flake8 pylint
 flake8:
 	flake8 .
 
+unit-test:
+	pytest tools/
+
 autopep8:
 	autopep8 --recursive --in-place .
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,3 +19,4 @@ bugzilla-data==0.0.5
 matplotlib==3.5.1
 sh==1.14.2
 semver==2.13.0
+pytest==7.1.1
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -13,6 +13,9 @@ lint: flake8 pylint
 flake8:
 	flake8 .
 
+unit-test: lint
```
osherdp, line 16: I think it's better to have unit-test independent from lint, as one's not always want to test a "clean" code

osherdp (changes requested):
Thanks for handling that!
Just one little comment

osherdp (approved):

PR #206, opened 2022-05-23
install python-bugzilla instead of the bigger scoped dependency

osherdp:
Seems like we're installing a much bigger-scoped dependency than intended, with not only API to bugzilla but also some matplotlib logic (?).

This change aims to fix that (and a couple of typos as well).


osherdp:
/?

PR #216, opened 2022-06-02
Raise a more graceful error when fields are None

osherdp:
Currently, whenever fields in Jira tickets are not being filled-in (e.g.
user / email domain fields), we [get the following error](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-assisted-assisted-installer-deployment-master-triage-status-report/1532255951992655872):
```
Traceback (most recent call last):
  File
"/home/assisted-installer-deployment/./tools/triage_status_report.py",
line 89, in <module>
    main()
  File
"/home/assisted-installer-deployment/./tools/triage_status_report.py",
line 85, in main
    triage_status_report(client, args.filter_id, args.webhook)
  File
"/home/assisted-installer-deployment/./tools/triage_status_report.py",
line 65, in triage_status_report
    for issue in sorted(_get_issues_data(issues)):
  File "<string>", line 4, in __lt__
TypeError: '<' not supported between instances of 'NoneType' and 'str'
```

Error is very hard to understand, so this change aims to validate fields
before dataclass logic for comparison actions tries to compare None
value with strings.

In addition, I've added a small message to let the user know when a slack message has been sent (currently it prints nothing which seems very odd)

/cc @omertuc

### DIFF
--- a/tools/triage_status_report.py
+++ b/tools/triage_status_report.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 import os
+import sys
 import json
 import argparse
 import dataclasses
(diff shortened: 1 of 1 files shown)


tools/triage_status_report.py
```diff
@@ -22,6 +22,11 @@ class IssueData:
     url: str
     features: List[str]
 
+    def __post_init__(self):
+        for field, value in self.__dict__.items():
+            if value is None:
+                raise ValueError(f"Field '{field}' cannot be None")
```
omertuc, line 28: Can we maybe collect all those exceptions and show them as part of the Slack message, instead of giving up completely? Something like 

"Here are the triage tickets:

And here are the triage tickets that could not be parsed: 

"

osherdp:
@omertuc when you have time, PTAL

PR #217, opened 2022-06-07
Fix case of no available RHCOS updates

osherdp:
Currently OCP 4.11 has no available candidates for the RHCOS images in "mirror.openshift.com". This leads to
``get_latest_rhcos_release_from_minor`` returning ``None``, and code doesn't handle that case.

/cc @eliorerz @danielerez 
/hold

### DIFF
--- a/tools/bump_ocp_releases.py
+++ b/tools/bump_ocp_releases.py
@@ -310,28 +310,30 @@ def update_os_images_json(default_os_images_json, updates_made, updates_made_str
 
         rhcos_latest_release = get_latest_rhcos_release_from_minor(openshift_version, rhcos_latest_of_releases, pre_release)
 
-        if rhcos_default_release != rhcos_latest_release:
-            updates_made.add(openshift_version)
-            updates_made_str.add(f"rhcos {rhcos_default_release} -> {rhcos_latest_release}")
+        if rhcos_latest_release is None or rhcos_default_release == rhcos_latest_release:
+            continue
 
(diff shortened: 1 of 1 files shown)


tools/bump_ocp_releases.py
```diff
@@ -310,28 +310,30 @@ def update_os_images_json(default_os_images_json, updates_made, updates_made_str
 
         rhcos_latest_release = get_latest_rhcos_release_from_minor(openshift_version, rhcos_latest_of_releases, pre_release)
 
-        if rhcos_default_release != rhcos_latest_release:
```
danielerez, line 313: can you check which releases are fetched in line 308? it should get all pre-releases: https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/pre-release/
and in line 311 it should get the nightly: https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/pre-release/4.11.0-0.nightly-2022-04-16-163450/

osherdp: Yeah, but it seems like we're filtering the nightly images: https://github.com/openshift-assisted/assisted-installer-deployment/blob/1cb3b5931044c0010cce9180974b818e41f378d4/tools/bump_ocp_releases.py#L163

I think for an okay reason. this script is mostly for promoting to tagged rhcos releases. I don't know why we don't have a tagged rhcos release for 4.11.0-fc.0 but eventually we should have one

PR #219, opened 2022-06-08
Add 'release' makefile target

osherdp:
For easier use of the ``release`` action (and to be able to chain
several targets in one command) this change adds this kind of target
that add git-tags to the relevant commits.


osherdp:
/unhold
(working
/cc @gamli75

gamli75 (approved):

PR #220, opened 2022-06-16
MGMT-10715: Skip creation of triage tickets for OCP 4.11

osherdp:
Since there's no available value for OCP 4.11 in "Affected Version"
field in Jira, and that we only have a fc version for it, we should
not create tickets for this version (for now).

/cc @omertuc @mkowalski @romfreiman

### DIFF
--- a/tools/create_triage_tickets.py
+++ b/tools/create_triage_tickets.py
@@ -19,6 +19,7 @@ import consts
 
 DEFAULT_DAYS_TO_HANDLE = 30
 DEFAULT_WATCHERS = ["mkowalsk"]
+SKIPPED_OCP_RELEASES = ["4.11"]
 
 
 LOGS_COLLECTOR = "http://assisted-logs-collector.usersys.redhat.com"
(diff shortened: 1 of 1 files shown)


omertuc:
It's a shame we wouldn't create tickets for 4.11 - it's interesting to find 4.11 / assisted compatibility bugs before 4.11 goes GA

omertuc:
>Since there's no available value for OCP 4.11 in "Affected Version"
field in Jira

Can we somehow make sure there's available values for 4.11, 4.12, 4.13, 4.14, etc all the way to 50 or something just so we won't have to deal with this limitation in the future?

osherdp:
> > Since there's no available value for OCP 4.11 in "Affected Version"
> > field in Jira
> 
> Can we somehow make sure there's available values for 4.11, 4.12, 4.13, 4.14, etc all the way to 50 or something just so we won't have to deal with this limitation in the future?

apparently I have permissions to do it in the ``AITRIAGE`` project, I got confused with ``MGMT``
so no need for this change, I'll add a couple of releases to it
/close

PR #221, opened 2022-06-20
add_triage_signature.py: autoclose nissho-ele.co.jp tickets

vrutkovs:
These customers are testing Contrail which has known issues on SDN side


osherdp:
/ok-to-test
FYI /cc @mkowalski

osherdp:
/cc @mkowalski
(FYI)

PR #227, opened 2022-06-22
Fix MediaDisconnection signature

danielerez:
The MediaDisconnectionSignature currently looks for the string 'media was likely disconnected' as an indication to include the signature.

But following https://github.com/openshift/assisted-service/pull/3124, the media disconnection message was [changed](https://github.com/openshift/assisted-service/blob/4b0c4757f3032ea6038434ea55534cb6bedbcbdb/internal/host/common.go#L21) and that exact string was removed. Thus, modified the ERRORS_PATTERNS string in the signature to: 'was either disconnected'.

Note: tested the change manually on: https://issues.redhat.com/browse/AITRIAGE-3342 (./tools/add_triage_signature.py -d -i AITRIAGE-3342)


omertuc:
Probably not too critical but would it make sense to look for both the old and new string?

omertuc:
I think not, so feel free to unhold if you agree

tools/add_triage_signature.py
```diff
@@ -878,7 +878,7 @@ class MediaDisconnectionSignature(ErrorSignature):
     Eventually, the signature will include common i/o errors and the number of times they appeared
     '''
 
-    ERRORS_PATTERNS = 'media was likely disconnected'
+    ERRORS_PATTERNS = 'was either disconnected'
```
osherdp, line 881: Maybe we should look for ``Unable to read from the discovery media`` instead?

danielerez:
> I think not, so feel free to unhold if you agree

Yeah it was changed more than half a year ago, so should be fine.

tools/add_triage_signature.py
danielerez: sure, makes sense.

mkowalski (approved):
Thanks!

osherdp (approved):

PR #228, opened 2022-06-22
signatures: update MediaDisconnectionSignature to match several patterns

vrutkovs:
Expand `ERRORS_PATTERNS` to match the host if it contains any of matching patterns.

This should help us catch issues like AITRIAGE-3352 earlier


omertuc:
https://github.com/openshift-assisted/assisted-installer-deployment/pull/227

tools/add_triage_signature.py
```diff
@@ -878,7 +878,11 @@ class MediaDisconnectionSignature(ErrorSignature):
     Eventually, the signature will include common i/o errors and the number of times they appeared
     '''
 
-    ERRORS_PATTERNS = 'Unable to read from the discovery media'
+    ERRORS_PATTERNS = [
+        'media was likely disconnected',
+        'Unable to read from the discovery media',
+        'Error: decoding and writing image',
```
omertuc, line 884: The third one indicates the user is faced with a cryptic coreos-installer image, it's not something we should disregard, it's something we should improve, so I don't feel it's fair to put in the same bucket as the first two

osherdp, line 884: I remotely remember something about it not being media disconnection but rather something else
@omertuc can you take a look?

osherdp (changes requested):

tools/add_triage_signature.py
mkowalski: +1 to Omer's comment; the last one is definitely not a clear user issue; it may be, but it may be not

vrutkovs: Hmm, should we rename the whole symptom into more generic "MediaIssues" then? Indicating that media has a problem on IO would save a lot of time during triage (even at the cost of being more vague)

omertuc: I just think we should sepearte "coreos-installer" errors from "assisted explicitly deciding we had media issues"

vrutkovs: OKay, reworking this to be a separate symptom

tools/add_triage_signature.py
```diff
@@ -911,6 +911,44 @@ def _update_ticket(self, url, issue_key, should_update=False):
             self._update_triaging_ticket(issue_key, report, should_update=should_update)
 
 
+
+class CoreOSInstallerErrorSignature(ErrorSignature):
+    '''
+    The signature looks for coreos-installer errors, usually due to bad disks
+    '''
+
+    ERROR_PATTERN = r'coreos-installer install .* Error exit status'
```
danielerez, line 920: Won't it catch also media disconnection issues? Maybe worth adding this signature only when 'Unable to read from the discovery media' is not in the status?

mkowalski (approved):

PR #229, opened 2022-06-22
Add reminder to upload logs as Jira attachments

omertuc:
Clusters don't remain on prod for too long - users reset them after a
while, or they get garbage collected. Sometimes the cluster on prod
has logs we don't have in Installation logs - if that's the case,
we should upload those logs as Jira attachments as soon as possible,
while we have that data.

This commit adds a bold red reminder under the clust on prod link to
remind triagers to do that


osherdp (approved):

mkowalski (approved):
Totally agree, I hope that works

PR #232, opened 2022-06-22
Update logs URL regex

omertuc:
Following 68ced8c8503b5f31434fae87fea33e54798d484f, the format of the
installation logs message has changed. The regex should be updated to
help it find the logs URL so the logs can be downloaded.


mkowalski (changes requested):
stop

PR #240, opened 2022-06-24
Copy "download_jobs" Jenkinsfile from test-infra repository

mkowalski:
/cc @osherdp 
/cc @adriengentil

### DIFF
--- a/Jenkinsfile.download_logs
+++ b/Jenkinsfile.download_logs
@@ -1,0 +1,67 @@
+pipeline {
+    agent { label 'download_logs' }
+
+    parameters {
+        string(name: 'REMOTE_SERVICE_URL', defaultValue: 'https://api.openshift.com', description: 'Service URL')
+        string(name: 'ASSISTED_INSTALLER_DEPLOYMENT_BRANCH', defaultValue: 'master', description: 'Branch on assisted-installer-deployment repo to use')
+        string(name: 'ASSISTED_TEST_INFRA_BRANCH', defaultValue: 'master', description: 'Branch on assisted-test-infra repo to use')
+    }
+
+    triggers { cron('H/30 * * * *') }
+
+    environment {
+        SKIPPER_PARAMS = " "
+
+        LOGS_DEST = "/var/ai-logs/triage_logs"
+
+        // Credentials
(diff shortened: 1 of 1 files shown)


Jenkinsfile.download_logs
```diff
@@ -0,0 +1,60 @@
...
+        stage('Cluster-logs') {
+            steps {
+                sh "make download_cluster_logs"
+            }
+        }
+
+        stage('Create tickets') {
+            steps {
+                sh "rm -rf assisted-installer-deployment"
+                sh "git clone https://github.com/openshift-assisted/assisted-installer-deployment --branch ${ASSISTED_INSTALLER_DEPLOYMENT_BRANCH}"
```
osherdp, line 42: You'll need to reverse that
Close test-infra and use assisted-installer-deployment

osherdp (changes requested):

Jenkinsfile.download_logs
osherdp: Sorry **clone

mkowalski: Correct, I did a clone of test-infra up in Init. In here I want to keep this clone to have an option to use job definition from master, but python scripts from a selected branch. What do you think?

Jenkinsfile.download_logs
```diff
@@ -0,0 +1,67 @@
...
+    }
+    options {
+      timeout(time: 90, unit: 'MINUTES')
+    }
+
+    stages {
+        stage('Init') {
+            steps {
+                sh "rm -rf assisted-test-infra"
+                sh "git clone https://github.com/openshift-assisted/assisted-test-infra --branch ${ASSISTED_TEST_INFRA_BRANCH}"
```
osherdp, line 30: You'll need to fix this to openshift instead of openshift-assisted

Jenkinsfile.download_logs
osherdp: Sounds good

osherdp (changes requested):

Jenkinsfile.download_logs
mkowalski: Of course, copy-paste mistake

osherdp (approved):

PR #242, opened 2022-06-24
Added new OSInstallationTime signature for hosts that took a long time to write RHCOS to disk

omertuc:
Looks like this:
image

For each host in the cluster, finds the:

```python
r'.*reached installation stage Writing image to disk$'
```

and

```python
r'.*reached installation stage Writing image to disk: 100%$'
```

events, takes the difference between their timestamps, and if it goes
over the threshold (300 seconds), it adds a warning about that
host which says how long writing the image to disk took.

If there's at-least one slow host, a table will be generated for all
hosts, showing the duration of all of them, even if they are below the
threshold.

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -109,8 +109,8 @@ def get_metadata_json(cluster_url):
 
 
 @functools.lru_cache(maxsize=1000)
-def get_events_json(cluster_url, cluster_id):
-    res = requests.get(f"{cluster_url}/cluster_{cluster_id}_events.json")
+def get_events_json(logs_url, cluster_id):
+    res = requests.get(f"{logs_url}/cluster_{cluster_id}_events.json")
     res.raise_for_status()
     return res.json()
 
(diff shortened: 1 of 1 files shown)


tools/add_triage_signature.py
```diff
@@ -1120,6 +1120,96 @@ def _update_ticket(self, url, issue_key, should_update=False):
...
+        events_by_host = defaultdict(list)
+        for event in get_events_json(cluster_url=url, cluster_id=cluster_id):
+            if "host_id" not in event:
+                # Cluster-level event, no host_id associated with it
+                continue
+            events_by_host[event["host_id"]].append(event)
+
+        hosts = []
+        for host in cluster['hosts']:
+            host_events = events_by_host[host["id"]]
```
mkowalski, line 1179: I wonder if we could have some generic function `get_events_by_host()` that would be accessible from different classes? That way if any other signature wants to parse events if will not need to have this logic of for-loop over all the events and building the map

tools/add_triage_signature.py
```diff
@@ -1120,6 +1120,96 @@ def _update_ticket(self, url, issue_key, should_update=False):
...
+
+            if total_duration_seconds <= 0:
+                # Negative writing to disk duration. This could happen if writing
+                # to disk never completed in this ticket but it did complete in
+                # previous runs.
+                return
+
+            if total_duration_seconds > self.slow_threshold_seconds:
+                host_disk_duration_seconds_message = (
+                    f"{{color:red}}OS installation to disk was rather slow, it took {total_duration_seconds} seconds{{color}}"
```
mkowalski, line 1198: I wonder if we could have the following logic (but treat it as a suggestion, I'm perfectly fine if you don't follow with it)

* if every host in the cluster installed faster than `slow_threshold_seconds`, don't print anything
* if at least one host in the cluster installed slower than `slow_threshold_seconds`, print `total_duration_seconds`

The latter is because if that happened that e.g. 1 out of 3 hosts exceeded a threshold, I would be interested to see what were values for the other hosts. For various reasons the following 2 scenarios are different

* 1 host installed in 15 minutes; 2 hosts installed in 1 minute
* 1 host installed in 5.5 minute; 2 hosts installed in 4.5 minute

mkowalski (approved):
Looks cool now, thank you!

PR #247, opened 2022-06-28
Add new NonstandardNetworkType signature and some refactoring

omertuc:
Sorry for the large diff

All these changes should make the script easier to maintain and
add code to in the future.

- Added NonstandardNetworkType
- Fixed bug where ErrorSignatures updated tickets even in dry run mode
- Formatted file with Black (created pyproject.toml with black settings)
- Fixed flake8 configuration to match black (in setup.cfg)
- Fixed pylint configuration to use a modern configuration file that
  matches the version we use (generated with `pylint --generate-rcfile > .pylintrc`)
- Use pip-tools (pip-compile) to render requirement.txt and
  dev-requirements.txt from *.in files, ensuring dependencies are
  frozen.
- Changed CF_* variables to more descriptive CUSTOM_FIELD_* name
- ErrorSignature's `_add_signature_label` method renamed to more correct `_add_signature_function_impact`
as it was modifying a custom field called function impact and not actual
labels. Also simplified its code a little bit
- Added an ability for ErrorSignature to modify actual Jira labels (as we're
  not sure about this function impact thing and whether it's actually
  used)
- Renamed SIGNATUERS to ALL_SIGNATURES
- Renamed should_update to should_reevaluate
- Child signatures are no longer aware of boring details like
  should_reevaluate (previously should_update), jira_client etc
- Child signatures no longer have to convert the log url format to the
  API form, it's now passed to them in the correct form


tools/add_triage_signature.py
```diff
@@ -189,94 +196,126 @@ def get_mustgather(triage_logs_tar):
...
+
+    def _add_signature_labels(self, issue_key, labels_to_add):
+        ticket_existing_labels = self._jira_client.issue(issue_key).fields.__dict__[FIELD_LABELS] or []
+        new_labels = [label for label in labels_to_add if label not in ticket_existing_labels]
+
+        if len(new_labels) != 0:
+            self._update_fields(
+                issue_key,
+                {FIELD_LABELS: ticket_existing_labels + new_labels},
+            )
```
mkowalski, line 256: I think this could be merged into one function that would accept an additional parameter, in this case `FIELD_LABELS` or `custom_field_name(CUSTOM_FIELD_FUNCTION_IMPACT)`. But feel free to keep 2 separate if you think it's easier to digest

tools/add_triage_signature.py
```diff
@@ -1010,45 +1167,53 @@ class ConsoleTimeoutSignature(ErrorSignature):
...
+                    [
+                        "h2. Waiting for console timeout probably due to VMware hosts on OCP 4.7",
+                        "You can mark it as caused by issue [MGMT-4454|https://issues.redhat.com/browse/MGMT-4454]",
+                        "Solution will be provided via [bugzilla|https://bugzilla.redhat.com/1935539] ticket",
+                    ]
+                )
 
         if report != "":
-            self._update_triaging_ticket(issue_key, report, should_update=should_update)
+            self._update_triaging_ticket(report)
```
mkowalski, line 1216: This one is so old... Can we maybe already remove it?

omertuc: Seems like its BZ was resolved so sure

vrutkovs (approved):
/lgtm
/hold

for another review

mkowalski:
I assume 2 people are enough to move forward with it. Let's merge

PR #258, opened 2022-06-30
Add hosts summary to HostsStatusSignature

vrutkovs:
Host status table may be confising. This changes a shorter summary about the state, i.e. "node pulled Ignition but didn't start kubelet" when status is Configuring.

Example for [AITRIAGE-3440](https://issues.redhat.com//browse/AITRIAGE-3440):

```
h2. Hosts status
|| id                                   || hostname   || progress                  || status   || role      || boot_mode   || status_info                                                                                                      || logs_info   || last_checked_in_at   ||
| e87d4d56-f7bb-0c89-22c3-95519cffdba6 | master03   | Waiting for control plane | error    | bootstrap | bios        | Host failed to install because its installation stage Waiting for control plane took longer than expected 1h0m0s | requested   | 2022-06-30 08:23:01  |
| 68334d56-f48c-25e2-d013-d05b8a1df7b0 | master02   | Configuring               | error    | master    | bios        | Host failed to install because its installation stage Configuring took longer than expected 1h0m0s               | completed   | 2022-06-30 07:15:33  |
| 49824d56-da15-a4d1-5cf6-9ba179813660 | master01   | Done                      | error    | master    | bios        | Host is part of a cluster that failed to install                                                                 | completed   | 2022-06-30 07:16:38  |
|| hostname   || role   || description                                     ||
| master02   | master | Node pulled Ignition, but never started kubelet |
```

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -345,6 +345,33 @@ class Signature(abc.ABC):
         inventory = json.loads(host["inventory"])
         return inventory["hostname"]
 
+    @staticmethod
+    def _generate_hosts_summary(hosts):
+        host_summary = []
+        for host in hosts:
+            # Add host status summary
+            if host["role"] not in ["master", "boostrap"]:
+                continue
+            if host["progress"] == "Done" or host["status"] != "error":
+                continue
+
+            host_comment = {
+                "Waiting for bootkube": "bootkube.service never completed",
+                "Rebooting": "Node never pulled Ignition",
+                "Configuring": "Node pulled Ignition, but never started kubelet",
+                "Waiting for control plane": "Masters never formed 2-node cluster",
+                "Waiting for controller": "Assisted installer controller pod never started",
+                "Writing image to disk": "Image probably failed to be written on disk",
+            }.get(host["progress"], "Unknown")
+            host_summary.append(
+                OrderedDict(
+                    hostname=host["hostname"],
+                    role=host["role"],
+                    description=host_comment,
+                )
+            )
+        return tabulate(host_summary, headers="keys", tablefmt="jira") + "\n"
+
 
 class ErrorSignature(Signature, abc.ABC):
     """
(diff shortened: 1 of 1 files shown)


tools/add_triage_signature.py
```diff
@@ -345,6 +345,33 @@ def _get_hostname(host):
...
+        host_summary = []
+        for host in hosts:
+            # Add host status summary
+            if host["role"] not in ["master", "boostrap"]:
+                continue
+            if host["progress"] == "Done" or host["status"] != "error":
+                continue
+
+            host_comment = {
+                "Waiting for bootkube": "Bootkube.service never started",
```
omertuc, line 359: It means bootkube [didn't finish](https://github.com/openshift/assisted-installer/blob/c13498b0e12bbe4ca3c1eea5e677b75712bba78b/src/installer/installer.go#L582), not that it didn't start

tools/add_triage_signature.py
```diff
@@ -345,6 +345,33 @@ def _get_hostname(host):
...
+                "Rebooting": "Node never pulled Ignition",
+                "Configuring": "Node pulled Ignition, but never started kubelet",
+                "Waiting for control plane": "Masters never formed 2-node cluster",
+                "Waiting for controller": "Assisted installer controller pod never started",
+                "Writing image to disk": "Image failed to be written on disk",
```
omertuc, line 364: This is a bit misleading because sometimes a stage ends prematurely when the cluster goes into error. But a cluster can go into error due to circumstances outside those happening on the node with the stage.

e.g. if workers are in the middle of writing the image to disk, and suddenly the bootstrap node unexpectedly disconnects from the service, the AI cluster will go into the error state, and as a result of that transition all hosts within the cluster will also move to the error state. But their "stage" (stages are "sub-states" of the Installing state) will remain the same and appear in that table.  

```suggestion
                "Writing image to disk": "Image probably failed to be written on disk",
```

PR #259, opened 2022-07-01
Add FlappingValidations

vrutkovs:
This would event logs for repetitive message, helping us find hosts which have unstable disk/network/ntp connection.

Example: https://issues.redhat.com/browse/AITRIAGE-3409

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -14,7 +14,7 @@ import sys
 import tempfile
 import yaml
 import ipaddress
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict, defaultdict, Counter
 from datetime import datetime
 from enum import Flag, auto
 from textwrap import dedent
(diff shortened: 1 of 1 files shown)


tools/add_triage_signature.py
```diff
@@ -1499,6 +1499,46 @@ def _process_ticket(self, url, issue_key):
...
+    def _process_ticket(self, url, issue_key):
+        validation_state_changes = {}
+        cluster = get_metadata_json(url)["cluster"]
+        events_by_host = get_events_by_host(url, cluster["id"])
+        for event in list(itertools.chain(*events_by_host.values())):
+            if any([regexp.match(event["message"]) for regexp in self.regexps]):
+                if event["message"] not in validation_state_changes:
+                    validation_state_changes[event["message"]] = 1
+                else:
+                    validation_state_changes[event["message"]] += 1
```
omertuc, line 1525: I believe this is equivalent:

```suggestion
        validation_state_changes = defaultdict(int)
        for event in get_events_json(url, get_metadata_json(url)["cluster"]["id"]):
            if any(regexp.match(event["message"]) for regexp in self.regexps):
                validation_state_changes[event["message"]] += 1
```

omertuc: Or even better:

```suggestion
        validation_state_changes = collections.Counter(
            event["message"] for event in
            get_events_json(url, get_metadata_json(url)["cluster"]["id"])
            if any(regexp.match(event["message"]) for regexp in self.regexps)
        )
```

vrutkovs: Good idea, implemented that

omertuc:
Nice idea

PR #264, opened 2022-07-10
Fix RHCOS image comparison and remove usage of ``rootfs_url``

osherdp:
1. We currently have problem comparing versions "4.11-fc.0" and "4.11.0-rc.0", because semver loose-version comparison transforms it to lists that might include either strings or integers, and then tries to compare all those values (leading to '<' not supported between instances of 'int' and 'str'). This change uses natsort which does not have this problem.
2. Remove update of the now absent "rootfs_url" key.

/cc @danielerez @eliorerz @carbonin 
/hold


tools/bump_ocp_releases.py
```diff
@@ -172,6 +189,7 @@ def get_latest_rhcos_release_from_minor(minor_release: str, all_releases: list,
     if not all_relevant_releases:
         return None
 
+    all_relevant_releases = list(get_canonized_rhcos_versions(all_relevant_releases))
```
danielerez, line 192: maybe worth trying something like [natsort](https://pypi.org/project/natsort/) instead?
from a quick test, seems it supports such scenarios:
`print(natsorted(["4.11.0-rc.1", "4.11-fc.0", "4.11.0-0.nightly"]))`
`output: ['4.11-fc.0', '4.11.0-0.nightly', '4.11.0-rc.1']`

osherdp: Done!

PR #267, opened 2022-07-12
MGMT-11123: allow skipping CCX installations with build-arg

osherdp:
Image currently builds with errors output, which makes for a red-herring when trying to find actual root-cause problems.

After this change gets in, we'll change CI image building to use that build-arg and skip performing CCX installation.
By default it should build with CCX, but even on failures it should not block on it (in case local env is not connected to VPN).

/cc @mkowalski @eliorerz 
/hold


Dockerfile.build
```diff
@@ -1,27 +0,0 @@
-FROM quay.io/centos/centos:stream8
```
osherdp, line 1: Apparently not used anymore, so I removed it

mkowalski:
/lgtm

Sounds reasonable, thanks!

osherdp:
works on jenkins as well: http://assisted-jenkins.usersys.redhat.com/job/download_logs/37734/console
/unhold

PR #271, opened 2022-07-19
Extract operator statuses from controller logs

omertuc:
The assisted controller dumps the status of all cluster operators into
its log. It's very hard to read for humans. This new signature will
parse those logs, find the unhealthy operators, and create a yaml
attachment that is more readable.

This is nice for when there is no must-gather but we do have
controller-logs. This information is also usually available
in Kraken, but no harm in parsing it from the controller as well.

You can see an example here:
https://issues.redhat.com/browse/AITRIAGE-3549?focusedCommentId=20651515&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20651515

Also fixed dry run bug where running, for example:
`./tools/add_triage_signature.py --issue [AITRIAGE-3549](https://issues.redhat.com//browse/AITRIAGE-3549) -us ControllerOperatorStatus`
would run in dry-run mode even though dry run is not specified.

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -1713,6 +1713,108 @@ class EventsInstallationAttempts(Signature):
             )
 
 
+class ControllerOperatorStatus(Signature):
+    """
+    This signature inspects the Assisted Controller logs to extract the status
+    of degraded operators and attaches them as a file in a human readable YAML
+    format
+    """
+
+    operator_regex = re.compile(r"Operator ([a-z\-]+), statuses: \[(.*)\].*")
+    conditions_regex = re.compile(r"\{(.+?)\}")
+    condition_regex = re.compile(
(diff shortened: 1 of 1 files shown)


tools/add_triage_signature.py
```diff
@@ -1713,6 +1713,108 @@ def _process_ticket(self, url, issue_key):
...
+            ),
+            aggregation_function=any,
+        )
+
+        if len(unhealthy_operators) != 0:
+            with tempfile.NamedTemporaryFile(
+                prefix="controller-logs-operators-status-", suffix=".yaml", mode="w"
+            ) as operator_status_tmp:
+                try:
+                    logger.debug(f"Writing Insights report as {operator_status_tmp.name}")
```
mkowalski, line 1802: Copy-pasta error, s/Insights report/operator status/

mkowalski (approved):

PR #275, opened 2022-07-27
NO-ISSUE: Stop releasing BZ tickets

adriengentil:
Since DPP-10714, we don't need to update BZ tickets anymore


adriengentil:
https://github.com/openshift-assisted/assisted-installer-deployment/pull/275/files#diff-025c14797ed9e1555da260e497ea4d095275f7818daafc5361d7c48d37d0af32R248-R250

Should be removed? I guess we want them in the release note, no? @osherdp

osherdp:
> https://github.com/openshift-assisted/assisted-installer-deployment/pull/275/files#diff-025c14797ed9e1555da260e497ea4d095275f7818daafc5361d7c48d37d0af32R248-R250
> 
> Should be removed? I guess we want them in the release note, no? @osherdp

Once there are no more BZ tickets with assisted-installer components, we most likely won't have PRs mentioning ``OCPBUGSM-*`` jira tickets (that are basically mirrors of the BZ ones). So iiuc you can remove this piece of code because either way we shouldn't have those cases in the longer term

osherdp:
btw, we are able now to remove python-bugzilla from requirements.txt?
I don't remember we use it for anything else

requirements.txt
```diff
@@ -10,7 +10,7 @@ attrs==21.4.0
     # via pytest
 beautifulsoup4==4.10.0
     # via -r requirements.in
-cachecontrol[filecache]==0.12.11
+cachecontrol[redis]==0.12.11
```
adriengentil, line 13: I see that this dependency is flapping: 
https://github.com/openshift-assisted/assisted-installer-deployment/pull/264/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R13

adriengentil:
> Once there are no more BZ tickets with assisted-installer components, we most likely won't have PRs mentioning `OCPBUGSM-*` jira tickets (that are basically mirrors of the BZ ones). So iiuc you can remove this piece of code because either way we shouldn't have those cases in the longer term

I remove all reference to `OCPBUGSM`



> btw, we are able now to remove python-bugzilla from requirements.txt? I don't remember we use it for anything else

I did a grep, and nothing else is using it, so I removed it

osherdp (approved):

PR #276, opened 2022-07-31
StaticNetworking signature

omertuc:
See example https://issues.redhat.com/browse/AITRIAGE-3677?focusedCommentId=20715928&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20715928

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -1847,6 +1847,35 @@ class DualStackBadRoute(Signature):
             )
 
 
+class StaticNetworking(Signature):
+    """
+    Shows the infraenv's static networking config
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            **kwargs,
+            comment_identifying_string="h1. At-least one infraenv has static networking config",
+        )
+
+    def _process_ticket(self, url, issue_key):
+        infraenvs = get_metadata_json(url).get("infraenvs", [])
+        messages = [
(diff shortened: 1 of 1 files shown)


romfreiman:
Nice!
Can u differentiate  if the statics networking was done by the UI template or user injecting the yaml? I think it will make a lot of difference. Also, if we we'll be able to identify Jiras according to it would be very usefull IMO.

Any chance you can find missconfigrations? Unsupported fields? Identation problems?

omertuc:
>Nice!
Can u differentiate if the statics networking was done by the UI template or user injecting the yaml? I think it will make a lot of difference. Also, if we we'll be able to identify Jiras according to it would be very usefull IMO.
Any chance you can find missconfigrations? Unsupported fields? Identation problems?

Would love to do all that in a separate PR. For now it's already useful in its current form so let's merge if there are no issues

eranco74 (commented):
/lgtm

eranco74:
> Nice! Can u differentiate if the statics networking was done by the UI template or user injecting the yaml? I think it will make a lot of difference. Also, if we we'll be able to identify Jiras according to it would be very usefull IMO.
> 
> Any chance you can find missconfigrations? Unsupported fields? Identation problems?

We can already identify Jiras according static-network-config
https://issues.redhat.com/issues/?jql=labels%20%3D%20FEATURE-Static-Network-Config

PR #278, opened 2022-08-03
Master failed rebooting

eranco74:
Add MasterFailedToPullIgnition signature, this signature finds tickets for clusters where the bootstrap node is `waiting for control plane` and both master nodes status is rebooting.
Fixed the signatures labeling
Updated triage_resolving_filters.json to close tickets with the above signature

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -249,7 +249,7 @@ class Signature(abc.ABC):
         self._identifing_string = comment_identifying_string
         self._old_identifing_string = old_comment_string
         self.dry_run_file = dry_run_file
-        self.should_reevaluate = dry_run_file
+        self.should_reevaluate = should_reevaluate
         self.issue_key = issue_key
 
     def process_ticket(self, url, issue_key):
@@ -283,10 +283,10 @@ class Signature(abc.ABC):
             )
 
     def _add_signature_function_impact(self, issue_key, labels_to_add):
-        self._add_labels_to_field(custom_field_name(CUSTOM_FIELD_FUNCTION_IMPACT))
+        self._add_labels_to_field(issue_key, labels_to_add, custom_field_name(CUSTOM_FIELD_FUNCTION_IMPACT))
 
     def _add_signature_labels(self, issue_key, labels_to_add):
-        self._add_labels_to_field(FIELD_LABELS)
+        self._add_labels_to_field(issue_key, labels_to_add, FIELD_LABELS)
 
     def find_signature_comment(self, key=None, comments=None):
         assert key or comments
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -714,6 +714,57 @@ class HostsExtraDetailSignature(Signature):
         self._update_triaging_ticket(report)
 
 
+class MasterFailedToPullIgnitionSignature(ErrorSignature):
+    """
+    This signature finds tickets for clusters where at least one master node failed to pull ignition
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            **kwargs,
+            comment_identifying_string="h1. Master nodes failed to pull ignition",
+            label="masters_failed_to_pull_ignition",
+        )
+
+    def _process_ticket(self, url, issue_key):
+        hosts = []
+        cluster_hosts = get_metadata_json(url)["cluster"]["hosts"]
+        # this signature is not relevant for SNO
+        if len(cluster_hosts) <= 1:
+            return
+
+        for host in cluster_hosts:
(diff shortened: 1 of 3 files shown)


tools/add_triage_signature.py
```diff
@@ -713,6 +713,49 @@ def _process_ticket(self, url, issue_key):
...
+                )
+        if len(hosts) > 0:
+            report = dedent(
+            """
+            When master node fail to come back from reboot there is nothing we can do.
+            """
+            )
+
+            report += self._generate_table_for_report(hosts)
+            self._update_triaging_ticket(report)
```
osherdp, line 759: nit: add another blank line

osherdp (changes requested):
Can you provide a screenshot for how it looks?
see ``./tools/add_triage_signature.py --help`` for how to do that on an example ticket

omertuc:
> Can you provide a screenshot for how it looks? see `./tools/add_triage_signature.py --help` for how to do that on an example ticket

Even better, run it without --dry-run and send us a link to the comment it generated

tools/add_triage_signature.py
```diff
@@ -713,6 +713,49 @@ def _process_ticket(self, url, issue_key):
...
+            **kwargs,
+            comment_identifying_string="h1. Master node failed to reboot",
+            function_impact_label="master_failed_rebooting",
+        )
+
+    def _process_ticket(self, url, issue_key):
+        md = get_metadata_json(url)
+        cluster = md["cluster"]
+        hosts = []
+        for host in cluster["hosts"]:
```
omertuc, line 734:
```suggestion
        hosts = []
        for host in get_metadata_json(url)["cluster"]["hosts"]:
```

tools/add_triage_signature.py
```diff
@@ -713,6 +713,49 @@ def _process_ticket(self, url, issue_key):
...
+            inventory = json.loads(host["inventory"])
+            boot_mode = inventory.get("boot", {}).get("current_boot_mode", "N/A")
+            if (
+                host["role"] == "master"
+                and host["progress"]["current_stage"] == "Rebooting"
+                and host["status"] == "error"
```
omertuc, line 740: It's also possible that the host went into error because the cluster went into error, can you prevent that from being a false positive?

tools/add_triage_signature.py
```diff
@@ -713,6 +713,49 @@ def _process_ticket(self, url, issue_key):
...
+                and boot_mode == "bios"
+            ):
+                hosts.append(
+                    OrderedDict(
+                        id=host["id"],
+                        hostname=inventory["hostname"],
+                        progress=host["progress"]["current_stage"],
```
omertuc, line 747: There appear in table titles so it's nicer if they're capitalized
```suggestion
                        HostID=host["id"],
                        Hostname=inventory["hostname"],
                        Progress=host["progress"]["current_stage"],
```

tools/add_triage_signature.py
```diff
@@ -713,6 +713,49 @@ def _process_ticket(self, url, issue_key):
...
+
+    def _process_ticket(self, url, issue_key):
+        md = get_metadata_json(url)
+        cluster = md["cluster"]
+        hosts = []
+        for host in cluster["hosts"]:
+            inventory = json.loads(host["inventory"])
+            boot_mode = inventory.get("boot", {}).get("current_boot_mode", "N/A")
+            if (
+                host["role"] == "master"
```
omertuc, line 738: Does this exclude bootstrap? Do we want to exclude bootstrap?

eranco74: It doesn't, in case a master node (with BIOS boot mode) failed to reboot we will label it

eranco74:
Screenshots:
Screenshot from 2022-08-03 14-36-43
Screenshot from 2022-08-03 14-35-59

tools/add_triage_signature.py
eranco74: Yes, I can check:
1. The bootstrap node status is waiting for control plane (in case it's not SNO).
2. If the master is in error than the current_stage must be rebooting
This should be unique enough for the case where the installation failed due to master reboot

tools/add_triage_signature.py
eranco74: Updating the logic here

eranco74:
> > Can you provide a screenshot for how it looks? see `./tools/add_triage_signature.py --help` for how to do that on an example ticket
> 
> Even better, run it without --dry-run on a ticket and send us a link to the comment it generated
Screenshot from 2022-08-03 15-03-06
Screenshot from 2022-08-03 15-02-25

osherdp:
Looks good to me 
@omertuc ?

osherdp (approved):

PR #282, opened 2022-08-10
add signature for hosts IPs

rccrdpccl:
This adds hostname and IP relation, to save some time when looking at issues

image

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -1989,6 +1989,51 @@ class NodeStatus(Signature):
         self._update_triaging_ticket(report)
 
 
+class HostsInterfacesSignature(Signature):
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            **kwargs,
+            comment_identifying_string="h1. Host interfaces:",
+        )
+
+    def _process_ticket(self, url, issue_key):
+        md = get_metadata_json(url)
+
+        cluster = md["cluster"]
+
+        hosts = []
+        for host in cluster["hosts"]:
+            interfaces = self._get_interfaces(host)
+            hosts.append(
+                OrderedDict(
+                    id=host["id"],
+                    hostname=self._get_hostname(host),
+                    name="\n".join(interfaces["name"]),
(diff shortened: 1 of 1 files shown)


tools/add_triage_signature.py
```diff
@@ -1989,6 +1989,47 @@ def _process_ticket(self, url, issue_key):
         self._update_triaging_ticket(report)
 
 
+class HostsIPSignature(Signature):
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            **kwargs,
+            comment_identifying_string="h1. Host IPv4 addresses:",
```
omertuc, line 1997: Why just IPv4?

tools/add_triage_signature.py
```diff
@@ -1989,6 +1989,47 @@ def _process_ticket(self, url, issue_key):
...
+
+    def _get_interfaces(self, inventory: dict):
+        ifaces = []
+        for interface in inventory.get("interfaces", []):
+            name = interface.get("name")
+            if not name:
+                continue
+            ifaces.append({
+                "name": name,
+                "ipv4_addresses": interface.get("ipv4_addresses", [])
```
omertuc, line 2027: We can have one table row per interface, see "Host storage details" signature for how it does it

omertuc: image
Right now it's not too readable

omertuc: Right now the ipv4_addresses column is not too readable

tools/add_triage_signature.py
```diff
@@ -1989,6 +1989,47 @@ def _process_ticket(self, url, issue_key):
...
+        cluster = md["cluster"]
+
+        hosts = []
+        for host in cluster["hosts"]:
+            inventory = json.loads(host["inventory"])
+            hosts.append(
+                OrderedDict(
+                    id=host["id"],
+                    hostname=self._get_hostname(host),
+                    ipv4_addresses=self._get_interfaces(inventory)
```
omertuc, line 2012: We should have one table row per interface, see "Host storage details" signature for how it does a similar thing. As of now ipv4_addresses is not readable enough

PR #289, opened 2022-08-19
Clarify static_networking message and make it more easily readable

javipolo:
/cc @omertuc


omertuc:
It's not necessarily just one host though

PR #292, opened 2022-08-23
MGMT-8890: Make bump_ocp_releases safe for missing data

mkowalski:
Contributes-to: MGMT-8890
Contributes-to: MGMT-11553


mkowalski:
/cc @osherdp 

Once we agree on how to nicely handle https://coreos.slack.com/archives/CFFJUNP6C/p1661241479471729 we can implement, but I think that nevertheless handling of `None` values is useful around

osherdp (approved):

PR #296, opened 2022-09-06
Don't fail on repos without base_ref

empovit:
In some cases (although it's not clear when), Prow job statuses may contain entries without `base_ref`. 
Let's ignore such jobs when we're not interested in them, and show a better message if the key is required.

### DIFF
--- a/tools/ci_status_report.py
+++ b/tools/ci_status_report.py
@@ -85,8 +85,10 @@ def filter_jobs(response: requests.Response) -> typing.Iterator[Job]:
         for ref in all_refs:
             organization = ref["org"]
             repository = ref["repo"]
-            base_ref = ref["base_ref"]
             if (organization, repository) in TRACKED_REPOSITORIES:
+                base_ref = ref.get("base_ref")
+                if not base_ref:
+                    raise Exception(f"Entry for repository {organization}/{repository} does not contain a 'base_ref' key")
                 break
 
         else:


tools/ci_status_report.py
```diff
@@ -85,8 +85,10 @@ def filter_jobs(response: requests.Response) -> typing.Iterator[Job]:
         for ref in all_refs:
             organization = ref["org"]
             repository = ref["repo"]
-            base_ref = ref["base_ref"]
             if (organization, repository) in TRACKED_REPOSITORIES:
+                base_ref = ref.get("base_ref")
+                if not base_ref:
+                    raise Exception(f"Entry for repository {organization}/{repository} does not contain a 'base_ref' key")
```
osherdp, line 91: Raise a more specific exception like ``ValueError`` for example

tools/ci_status_report.py
```diff
@@ -85,8 +85,10 @@ def filter_jobs(response: requests.Response) -> typing.Iterator[Job]:
         for ref in all_refs:
             organization = ref["org"]
             repository = ref["repo"]
-            base_ref = ref["base_ref"]
             if (organization, repository) in TRACKED_REPOSITORIES:
+                base_ref = ref.get("base_ref")
+                if not base_ref:
```
osherdp, line 90: wouldn't it still raise an exception?
I mean, in case it's not available in the dictionary ``base_ref`` will be ``None`` which still gets into raising the exception

osherdp (changes requested):

tools/ci_status_report.py
osherdp: either way the ``ci-status-report-dry-run`` job will answer that question

osherdp: got it. it will only apply to our repos
that's good

osherdp (approved):

PR #297, opened 2022-09-09
MGMT-11940: Add signature for user_managed_networking and load-balancer issues

javipolo:
/cc @omertuc
/cc @mkowalski

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -192,6 +192,12 @@ def get_triage_logs_tar(triage_url, cluster_id):
         raise FailedToGetLogsTarException from e
 
 
+def get_controller_logs(triage_url):
+    return get_triage_logs_tar(triage_url=triage_url, cluster_id=get_metadata_json(triage_url)["cluster"]["id"]).get(
+        "controller_logs.tar.gz/assisted-installer-controller-*.logs"
+    )
+
+
 def get_host_log_file(triage_logs_tar, host_id, filename):
     # The file is already uniquely determined by the host_id, we can omit the hostname
     hostname = "*"
@@ -668,9 +674,8 @@ class FailureDetails(Signature):
         #           in the https://github.com/openshift/assisted-service/blob/master/internal/usage/consts.go
         #           we are manually extracting it from the cluster config and setting a feature flag
         #           so that JIRA issues can be filtered already now.
-        if "user_managed_networking" in cluster_md:
-            if cluster_md["user_managed_networking"] is True or cluster_md["user_managed_networking"] == "true":
-                update_fields["labels"].append("FEATURE-User-Managed-Networking")
+        if cluster_md.get("user_managed_networking", False):
+            update_fields["labels"].append("FEATURE-User-Managed-Networking")
 
         # (mko) Platform is not implemented as a separate feature flag, but we want to be able to filter
         #       it in Jira using labels. For this reason we are extracting it manually from the cluster payload.
@@ -1182,14 +1187,8 @@ class ApiInvalidCertificateSignature(ErrorSignature):
         )
 
     def _process_ticket(self, url, issue_key):
-        md = get_metadata_json(url)
-
-        cluster = md["cluster"]
-        cluster_id = cluster["id"]
-        triage_logs_tar = get_triage_logs_tar(triage_url=url, cluster_id=cluster_id)
-
         try:
-            controller_logs = triage_logs_tar.get("controller_logs.tar.gz/assisted-installer-controller-*.logs")
+            controller_logs = get_controller_logs(url)
         except FileNotFoundError:
             return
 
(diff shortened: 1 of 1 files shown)


tools/add_triage_signature.py
```diff
@@ -2149,6 +2124,56 @@ def _process_ticket(self, url, issue_key):
...
 
+class UserManagedNetworkingLoadBalancer(ErrorSignature):
+    lb_operators = ['authentication', 'console', 'ingress']
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            **kwargs,
+            function_impact_label="umh_load_balancer_unhealthy",
+            comment_identifying_string="h1. Probably user managed load-balancer issues, nothing we can do",
```
omertuc, line 2135:
```suggestion
            comment_identifying_string="h1. Probably user managed load-balancer issues, <could theoretically be prevented with MGMT-TheTicketWithTheLoadBalancerValidationThatYouCreatedYesterday",
```

tools/add_triage_signature.py
```diff
@@ -2149,6 +2124,56 @@ def _process_ticket(self, url, issue_key):
...
+
+        try:
+            controller_logs = get_triage_logs_tar(
+                triage_url=url, cluster_id=get_metadata_json(url)["cluster"]["id"]
+            ).get("controller_logs.tar.gz/assisted-installer-controller-*.logs")
+        except FileNotFoundError:
+            return
+
+        statuses = operator_statuses_from_controller_logs(controller_logs)
+        unhealthy_operators = [operator for operator in statuses if self.operator_unhealthy(statuses[operator])]
```
omertuc, line 2162: You can also reuse filter_operators from the other sig

tools/add_triage_signature.py
```diff
@@ -2149,6 +2124,53 @@ def _process_ticket(self, url, issue_key):
...
+
+        if self.lb_operators.sort() == unhealthy_operators.sort():
```
omertuc, line 2161: Compare Python [sets](https://realpython.com/python-sets/) rather than lists, this way you can avoid sort() as order doesn't matter

tools/add_triage_signature.py
```diff
@@ -2149,6 +2124,53 @@ def _process_ticket(self, url, issue_key):
...
+        return False
+
+    def _process_ticket(self, url, issue_key):
+        md = get_metadata_json(url)
+        cluster_md = md["cluster"]
+        if cluster_md["user_managed_networking"] is not True and cluster_md["user_managed_networking"] != "true":
+            return
+
+        try:
+            controller_logs = get_triage_logs_tar(
```
omertuc, line 2152: Can we move this nasty call to a function reused by all signatures that have it?

omertuc: `get_controller_logs` or something

tools/add_triage_signature.py
```diff
@@ -2149,6 +2101,53 @@ def _process_ticket(self, url, issue_key):
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            **kwargs,
+            function_impact_label="umh_load_balancer_unhealthy",
+            comment_identifying_string="h1. Probably user managed load-balancer issues, could theoretically be prevented with MGMT-11941",
+        )
+
+    def operator_unhealthy(self, operator):
```
omertuc, line 2115: Unused

tools/add_triage_signature.py
```diff
@@ -29,6 +29,7 @@
 import tqdm
 from fuzzywuzzy import fuzz
 from tabulate import tabulate
+import pprint
```
omertuc, line 32: Unused

tools/add_triage_signature.py
javipolo: Done! :)

tools/add_triage_signature.py
```diff
@@ -2149,6 +2098,44 @@ def _process_ticket(self, url, issue_key):
...
+            *args,
+            **kwargs,
+            function_impact_label="umh_load_balancer_unhealthy",
+            comment_identifying_string="h1. Probably user managed load-balancer issues, could theoretically be prevented with MGMT-11941",
+        )
+
+    def _process_ticket(self, url, issue_key):
+        md = get_metadata_json(url)
+        cluster_md = md["cluster"]
+        if cluster_md["user_managed_networking"] is not True and cluster_md["user_managed_networking"] != "true":
```
omertuc, line 2115:
```suggestion
        if cluster_md.get("user_managed_networking", False):
```

omertuc: The value is a simple boolean, there's no reason to believe it will contain the string "false", did you see anything like that?

tools/add_triage_signature.py
```diff
@@ -2149,6 +2098,44 @@ def _process_ticket(self, url, issue_key):
             self._update_triaging_ticket("\n".join(host_messages))
 
 
+class UserManagedNetworkingLoadBalancer(ErrorSignature):
+    lb_operators = set(['authentication', 'console', 'ingress'])
```
omertuc, line 2102:
```suggestion
    lb_operators = {"authentication", "console", "ingress"}
```

omertuc: ({ } is a set literal, double quotes because we use the Black formatter and that's the quotes Black prefers)

omertuc:
Can you run it once without --dry-run on the ticket for which you tested it, and then link to the comment that was generated? Just so we can see a live example

omertuc:
Also CI in this repo is practically free so no need for draft PRs

tools/add_triage_signature.py
javipolo: I just copied the check from https://github.com/openshift-assisted/assisted-installer-deployment/blob/e7b869a3cff31dde659ee0d6fada0aca2c226289/tools/add_triage_signature.py#L672 , I dont know the reason to check both types ... maybe it's for backwards compatibility and it's not neede anymore :?

@mkowalski do you know about this?

javipolo:
See example: https://issues.redhat.com/browse/AITRIAGE-3950?focusedCommentId=20923638&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20923638

tools/add_triage_signature.py
```diff
@@ -2149,6 +2098,37 @@ def _process_ticket(self, url, issue_key):
...
+            controller_logs = get_controller_logs(url)
+        except FileNotFoundError:
+            return
+
+        unhealthy_operators = filter_operators(
+            operator_statuses_from_controller_logs(controller_logs),
+            (("Degraded", True), ("Available", False), ("Progressing", True)),
+            aggregation_function=any)
+
+        if self.lb_operators == set(unhealthy_operators.keys()):
```
omertuc, line 2128: You can compare dict keys to a set directly without converting the dict keys to a set

```[ins] In [7]: a = {1:2, 2:3}

[ins] In [8]: a.keys() == {1,2}
Out[8]: True

[ins] In [9]: a.keys() == {2,1}
Out[9]: True```

File: tools/add_triage_signature.py [hunk 10]
```diff
@@ -2149,6 +2098,37 @@ def _process_ticket(self, url, issue_key):
...
+
+        try:
+            controller_logs = get_controller_logs(url)
+        except FileNotFoundError:
+            return
+
+        unhealthy_operators = filter_operators(
+            operator_statuses_from_controller_logs(controller_logs),
+            (("Degraded", True), ("Available", False), ("Progressing", True)),
+            aggregation_function=any)
```
omertuc, line 2126: `black tools/add_triage_signature.py`

tools/add_triage_signature.py
mkowalski: Yeah, back then when I wrote this I tried to find an ultimate answer to “what is the type here” question, but given it’s Python without typing, I didn’t manage to. If anyone is perfectly sure we have here one type and not the other I’m fine with this

omertuc: Okay then let's go with my edit suggestion

PR #304, opened 2022-10-11
NO-ISSUE: Don't assign auto-closed tickets to a human

mkowalski:
/cc @adriengentil 
/cc @omertuc


omertuc:
Leacing

omertuc:
Leaving behind commented out code is unnecessary, we can always just revert this commit if we ever want it back

omertuc:
Sorry closed due to misclick

mkowalski:
Totally agree, just fixed and completely removed that

PR #308, opened 2022-10-17
MGMT-11875: Consider operators without conditions unhealthy

jhernand:
Currently the operators that don't have explicit conditions reported in the logs aren't consider unhealthy. For example, the following log lines won't result in adding the operators to the generated YAML files:

```
time="2022-08-31T03:47:06Z" level=info msg="Operator openshift-controller-manager, statuses: []"
time="2022-08-31T03:47:06Z" level=info msg="Operator openshift-samples, statuses: []"
```

This patch changes the `ControllerOperatorStatus` signature so that it will consider those operators unhealthy and generate a YAML file like this:

```
openshift-controller-manager: {}
openshift-samples: {}
```

Related: https://issues.redhat.com/browse/MGMT-11875

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -1717,8 +1717,12 @@ class ControllerOperatorStatus(Signature):
         except FileNotFoundError:
             return
 
+        # We want to consider unhealthy operators that have a condition explicitly marking them as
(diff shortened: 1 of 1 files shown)


omertuc:
Thank you!

PR #315, opened 2022-11-29
Explicit message for when there's no install-config

omertuc:
Example:

https://issues.redhat.com/browse/AITRIAGE-4631?focusedCommentId=21331871&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21331871

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -478,7 +478,15 @@ class HostsStatusSignature(Signature):
 
     def _process_ticket(self, url, issue_key):
         md = get_metadata_json(url)
-        installconfig = get_installconfig_yaml(url)
+
+        try:
+            installconfig = get_installconfig_yaml(url)
+        except FailedToGetInstallConfigException:
+            logger.exception("Failed to get install-config.yaml")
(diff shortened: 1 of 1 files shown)


tools/add_triage_signature.py
```diff
@@ -478,7 +478,14 @@ def __init__(self, *args, **kwargs):
...
     def _process_ticket(self, url, issue_key):
         md = get_metadata_json(url)
-        installconfig = get_installconfig_yaml(url)
+
+        try:
+            installconfig = get_installconfig_yaml(url)
+        except FailedToGetInstallConfigException:
+            self._update_triaging_ticket(
+                "Installation status signature cannot be created. This cluster's collected logs are missing install-config.yaml for an unknown reason, why is it missing?"
+            )
```
osherdp, line 487: maybe we should log the error as well
it will still not be easy to find the root-cause maybe (having to look throughout a couple of jenkins jobs logs) but at least we'll have it written somewhere

omertuc: Done

PR #317, opened 2022-12-08
MGMT-12867 Add a signature to nsenter failures

filanov:
Not all nsemter command will cause the agnet to fail command execution, This signature will help to catch those cases

an example:

level=info msg=\"failed executing nsenter [--target 1 --cgroup --mount --ipc --pid -- mdadm --stop /dev/md127]

the level of this command is info and it will not cause the agnet to fail the step execution but it did cause an issue with the installation later

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -428,7 +428,6 @@ class Signature(abc.ABC):
             )
         return tabulate(host_summary, headers="keys", tablefmt="jira") + "\n"
 
-
 class ErrorSignature(Signature, abc.ABC):
     """
     ErrorSignature is a Signature that also optionally adds function impact and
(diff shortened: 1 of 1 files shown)


filanov:
/hold
not sure how to test it

omertuc:
> > not sure how to test it
> 
> `tools/add_triage_signature.py` can run this check in dry run mode against a particular triage issue (`-i`) or recent (30 days) (`-r`)

See `CONTRIBUTING.md` for more details on signature testing / development

omertuc:
Yes I agree, this should not be "nsenter" specific. We use nsenter when we want to escape the container into the host's namespaces, but that's just an implementation detail that could theoretically change

filanov:
My reference was https://issues.redhat.com/browse/AITRIAGE-4689 
Looking at the ticket again i can see that step signature does cover this case. 
looking at the existing signature i couldn't see the failure this is why it looked like it wasn't covered. 
Closing this PR

PR #325, opened 2023-01-02
New ControllerWarnings signature to show controller warnings

omertuc:
The controller logs might contain interesting warnings we want to
bring to the attention of triagers

You can see an example here https://issues.redhat.com/browse/AITRIAGE-4861?focusedCommentId=21478853&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21478853

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -2295,6 +2295,38 @@ class FailedRequestTriggersHostTimeout(Signature):
         return {host["id"] for host in cluster["hosts"] if host["status_info"] == cls.HOST_TIMED_OUT_STATUS_INFO}
 
 
+class ControllerWarnings(Signature):
+    """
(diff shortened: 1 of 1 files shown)


tools/add_triage_signature.py
```diff
@@ -2295,6 +2295,35 @@ def _timed_out_hosts(cls, cluster):
...
+        super().__init__(
+            *args,
+            **kwargs,
+            comment_identifying_string="h1. Controller warning logs",
+        )
+
+    def _process_ticket(self, url, issue_key):
+        try:
+            controller_logs = get_controller_logs(url)
+        except FileNotFoundError:
```
osherdp, line 2313: let's add debug-log for those cases, just so we can track the flow

PR #326, opened 2023-01-04
temp workaround to must-gather sig

slaviered:
until a fix to logs_info timing is in prod

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -1394,7 +1394,18 @@ class MissingMustGatherLogs(ErrorSignature):
     def _process_ticket(self, url, issue_key):
         md = get_metadata_json(url)
         cluster_id = md["cluster"]["id"]
-        if md["cluster"]["logs_info"] in ("timeout", "completed"):
+        # temporary workaround till requested status is set only when the controller is
+        # already up and running
+        cluster = md["cluster"]
+        hosts = cluster["hosts"]
+        masters_in_done = len(host for host in hosts if host["progress"] == "Done")
+        def bootstrap_after_waiting_for_bootkube():
+            for host in hosts:
+                if host["role"] == "bootstrap" and host["progress"]["current_stage"] in ["Rebooting", "Configuring", "Joined", "Done"]:
+                    return True
+            return False
+        ##
+        if cluster["logs_info"] in ("timeout", "completed") and masters_in_done >= 2 and bootstrap_after_waiting_for_bootkube():
             triage_logs_tar = get_triage_logs_tar(triage_url=url, cluster_id=cluster_id)
             try:
                 get_mustgather(triage_logs_tar)


tools/add_triage_signature.py
```diff
@@ -1394,7 +1394,13 @@ def __init__(self, *args, **kwargs):
...
         md = get_metadata_json(url)
         cluster_id = md["cluster"]["id"]
-        if md["cluster"]["logs_info"] in ("timeout", "completed"):
+        # temporary workaround till requested status is set only when the controller is
+        # already up and running
+        cluster = md["cluster"]
+        hosts = cluster["hosts"]
+        some_done = any(host["progress"] == "Done" for host in hosts)
+        ##
+        if some_done and md["cluster"]["logs_info"] in ("timeout", "completed", "collecting"):
```
eranco74, line 1403: Why collecting?
if it's still collecting we might still get it

slaviered: We run it on failed clusters so that might be an edge case where the triage ticket is scrapped before the must-gather has finished (by mistake). This signature is designed to find such strange edge cases

eranco74: That cluster logs will keep getting updated until the logs_info reach `timeout` or `collected`. 
see https://github.com/openshift/assisted-test-infra/pull/1974/files

slaviered: collecting removed since the signature can run any time during the cluster lifecycle and may report inaccurate status for collecting state that may eventually result in successful log scrapping.

eranco74:
@tsorya I think you are working on code changes that should allow the controller to start running before the node status is set to Done.
Can you advise on what this code should look like to accommodate your changes?

slaviered:
@tsorya @eranco74 in any case this is only temporary until I am moving the logs_info=required to where the controller is already up and running

tools/add_triage_signature.py
```diff
@@ -1394,7 +1394,13 @@ def __init__(self, *args, **kwargs):
     def _process_ticket(self, url, issue_key):
         md = get_metadata_json(url)
         cluster_id = md["cluster"]["id"]
-        if md["cluster"]["logs_info"] in ("timeout", "completed"):
+        # temporary workaround till requested status is set only when the controller is
+        # already up and running
+        cluster = md["cluster"]
```
tsorya, line 1399: why not to check finalizing state? till then must-gather will not be available

tsorya: We need to build on must-gather only if 3 masters are installed till then i don't think it will work.

slaviered: @tsorya what does it mean 3 masters are installed? All in Done ?

tsorya: Yep that means  == finilizing state in cluster

eranco74: But @tsorya we should be able to get must-gather logs if 1 master is done, no?

eranco74:
replaced by
https://github.com/openshift-assisted/assisted-installer-deployment/pull/331

PR #331, opened 2023-01-15
workaround to must-gather sig

eranco74:
currently the signature show up in cases that we know that the must-gather can't be collected.
With this fix triage tickets will get this signature only if must-gather logs should've been collected - the bootstrap node went to reboot. In case of SNO the signature will be added only if we the controller start running post reboot.

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -1393,8 +1393,19 @@ class MissingMustGatherLogs(ErrorSignature):
 
     def _process_ticket(self, url, issue_key):
         md = get_metadata_json(url)
-        cluster_id = md["cluster"]["id"]
+        cluster_hosts = md["cluster"]["hosts"]
+        bootstrap_node = [host for host in cluster_hosts if host["bootstrap"]][0]
+
+        eligible_bootstrap_stages = ["Rebooting", "Configuring", "Joined", "Done"]
+        if len(cluster_hosts) <= 1:
+            # in SNO when the bootstrap node goes to reboot the controller is still not running
+            eligible_bootstrap_stages.remove("Rebooting")
+
+        if bootstrap_node["progress"]["current_stage"] not in eligible_bootstrap_stages:
+            return
+
         if md["cluster"]["logs_info"] in ("timeout", "completed"):
+            cluster_id = md["cluster"]["id"]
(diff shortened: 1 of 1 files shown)


tools/add_triage_signature.py
```diff
@@ -1393,8 +1393,16 @@ def __init__(self, *args, **kwargs):
 
     def _process_ticket(self, url, issue_key):
         md = get_metadata_json(url)
-        cluster_id = md["cluster"]["id"]
+        cluster_hosts = md["cluster"]["hosts"]
+        bootstrap_node = [host for host in cluster_hosts if host["bootstrap"]][0]
+        eligible_bootstrap_stages = ["Rebooting", "Configuring", "Joined", "Done"]
```
osherdp, line 1398: also ``Failed``, no?

tools/add_triage_signature.py
```diff
@@ -1393,8 +1393,16 @@ def __init__(self, *args, **kwargs):
...
+        if len(cluster_hosts) <= 1:
+            # in SNO when the bootstrap node goes to reboot the controller is still not running
+            eligible_bootstrap_stages.remove("Rebooting")
```
osherdp, line 1401: append blank line here, I'm choking here from all those lines XD

tools/add_triage_signature.py
```diff
@@ -1393,8 +1393,16 @@ def __init__(self, *args, **kwargs):
...
+        if bootstrap_node["progress"]["current_stage"] not in eligible_bootstrap_stages:
+            return
```
osherdp, line 1403: here too

osherdp (changes requested):
looks good. styling and a question

tools/add_triage_signature.py
eranco74: Nope, this is progress stage, not state

eranco74: Bootstrap progress stages are:
         "Starting installation",
          "Installing",
          "Writing image to disk",
          "Waiting for control plane",
          "Waiting for bootkube",
          "Waiting for controller",
          "Rebooting",
          "Configuring",
          "Joined",
          "Done"

osherdp (approved):

PR #332, opened 2023-01-15
Add ControllerFailedToStart signature

eranco74:
It seems that sometimes the installation fails because the installer pod fail to be ready.
This is the first version of this signature.
See example signature [here](https://issues.redhat.com/browse/AITRIAGE-4917)
It should be extended with the reason that prevented the pod from starting/reaching a ready status.

### DIFF
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -2399,6 +2399,74 @@ class MissingOSTreePivot(ErrorSignature):
             )
 
 
+class ControllerFailedToStart(ErrorSignature):
+    """
+    This signature looks for missing pivot URL in rpm-ostree status of the control-plane hosts
+    """
+
+    PODS_PATH = "*_bootstrap_*.tar.gz/logs_host_*/log-bundle-*.tar.gz/log-bundle-*/resources/pods.json"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            **kwargs,
+            function_impact_label="assisted_controller_failed_to_start",
+            comment_identifying_string="h1. Assisted Installer Controller failed to start",
+        )
+
+    def _process_ticket(self, url, issue_key):
+        cluster = get_metadata_json(url)["cluster"]
+        bootstrap_node = [host for host in cluster["hosts"] if host["bootstrap"]][0]
+
+        if bootstrap_node["progress"]["current_stage"] != "Waiting for controller":
+            return
+
+        triage_logs_tar = get_triage_logs_tar(triage_url=url, cluster_id=cluster["id"])
+        try:
+            # Fetch pods.json from the bootstrap log bundle
+            pods_json = triage_logs_tar.get(self.PODS_PATH)
+        except FileNotFoundError:
+            return
+
+        # Fetch assisted-controller status
+        try:
+            pods = json.loads(pods_json)
+            try:
+                controller_pod = [
+                    pod
+                    for pod in pods.get("items", [])
+                    if pod.get("metadata", {}).get("namespace") == "assisted-installer"
+                ][0]
+            except IndexError:
+                raise RuntimeError("Failed to find controller pod in pods json")
+            ready = False
+            try:
+                ready = [
+                    condition.get("status") == "True"
+                    for condition in controller_pod.get("status", {}).get("conditions", {})
+                    if condition.get("type") == "Ready"
+                ][0]
+            except IndexError:
+                raise RuntimeError(
+                    f"Failed to determine controller pod ready condition, here is it's status: {controller_pod['status']}"
+                )
+            self._update_triaging_ticket(
+                dedent(
+                    f"""
+            The controller pod {"is" if ready else "isn't"} ready, here is its status:
+            h3. conditions:
+            {self._generate_table_for_report(controller_pod["status"]["conditions"])}
+            h3. containerStatuses:
+            {self._generate_table_for_report(controller_pod["status"]["containerStatuses"])}
+            """
+                )
+            )
+        except Exception as e:
+            self._update_triaging_ticket(
+                f"Failed to get readiness of controller pod from installer-gather resources/pods.json: {e}"
+            )
+
+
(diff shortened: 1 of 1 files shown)


tools/add_triage_signature.py
```diff
@@ -2399,6 +2399,61 @@ def _process_ticket(self, url, issue_key):
...
+        try:
+            # Fetch pods.json from the bootstrap log bundle
+            pods_json = triage_logs_tar.get(self.PODS_PATH)
+        except FileNotFoundError:
+            return
+
+        # Fetch assisted-controller status
+        try:
+            pods = json.loads(pods_json)
+            controller_pod = [pod for pod in pods["items"] if pod["metadata"]["namespace"] == "assisted-installer"][0]
```
omertuc, line 2435: Let's make another signature that does a generic dump table of all unhappy pods in that JSON, and their containerStatuses? Might be more revealing of a widespread issue than looking at just our container

tools/add_triage_signature.py
```diff
@@ -2399,6 +2399,61 @@ def _process_ticket(self, url, issue_key):
...
+
+    def _process_ticket(self, url, issue_key):
+        md = get_metadata_json(url)
+        cluster_hosts = md["cluster"]["hosts"]
+        bootstrap_node = [host for host in cluster_hosts if host["bootstrap"]][0]
+
+        if bootstrap_node["progress"]["current_stage"] != "Waiting for controller":
+            return
+
+        triage_logs_tar = get_triage_logs_tar(triage_url=url, cluster_id=get_metadata_json(url)["cluster"]["id"])
```
omertuc, line 2425: Don't download metadata twice

```suggestion
        cluster = get_metadata_json(url)["cluster"]
        bootstrap_node = [host for host in cluster["hosts"] if host["bootstrap"]][0]

        if bootstrap_node["progress"]["current_stage"] != "Waiting for controller":
            return

        triage_logs_tar = get_triage_logs_tar(triage_url=url, cluster_id=cluster["id"])
```

tools/add_triage_signature.py
omertuc, line 2435: Let's have more graceful handling of the controller pod `[0]` not existing at all

tools/add_triage_signature.py
```diff
@@ -2399,6 +2399,61 @@ def _process_ticket(self, url, issue_key):
...
+            # Fetch pods.json from the bootstrap log bundle
+            pods_json = triage_logs_tar.get(self.PODS_PATH)
+        except FileNotFoundError:
+            return
+
+        # Fetch assisted-controller status
+        try:
+            pods = json.loads(pods_json)
+            controller_pod = [pod for pod in pods["items"] if pod["metadata"]["namespace"] == "assisted-installer"][0]
+            ready = [
```
omertuc, line 2436: Also more graceful handling of "status", "conditions", "type", and the Ready condition not being present

tools/add_triage_signature.py
```diff
@@ -2399,6 +2399,61 @@ def _process_ticket(self, url, issue_key):
...
+                The controller pod isn't ready, here is it's status:
+                h3. conditions:
+                {self._generate_table_for_report(controller_pod["status"]["conditions"])}
+                h3. containerStatuses:
+                {self._generate_table_for_report(controller_pod["status"]["containerStatuses"])}
+                """
+                    )
+                )
+        except Exception:
+            self._update_triaging_ticket("Failed to get controller pod from pods.json")
```
omertuc, line 2454:
```suggestion
        except Exception as e:
            self._update_triaging_ticket("Failed to get readiness of controller pod from installer-gather resources/pods.json: {e}")
```

Include the error in the sig

tools/add_triage_signature.py
eranco74: In another PR...

tools/add_triage_signature.py
```diff
@@ -2399,6 +2399,72 @@ def _process_ticket(self, url, issue_key):
...
+            except IndexError:
+                raise "Failed to find controller pod in pods json"
+            try:
+                ready = [
+                    condition.get("status") == "True"
+                    for condition in controller_pod.get("status", {}).get("conditions", {})
+                    if condition.get("type") == "Ready"
+                ][0]
+            except IndexError:
+                raise f"Failed to determine controller pod in ready condition, here is it's status: {controller_pod['status']}"
```
osherdp, line 2449:
```suggestion
                raise RuntimeError(f"Failed to determine controller pod in ready condition, here is it's status: {controller_pod['status']}")
```

tools/add_triage_signature.py
```diff
@@ -2399,6 +2399,72 @@ def _process_ticket(self, url, issue_key):
...
+            if not ready:
+                self._update_triaging_ticket(
+                    dedent(
+                        f"""
+                The controller pod isn't ready, here is it's status:
```
osherdp, line 2454:
```suggestion
                The controller pod isn't ready, here is its status:
```

osherdp (changes requested):

tools/add_triage_signature.py
```diff
@@ -2399,6 +2399,72 @@ def _process_ticket(self, url, issue_key):
...
+                raise "Failed to find controller pod in pods json"
+            try:
+                ready = [
+                    condition.get("status") == "True"
+                    for condition in controller_pod.get("status", {}).get("conditions", {})
+                    if condition.get("type") == "Ready"
+                ][0]
+            except IndexError:
+                raise f"Failed to determine controller pod in ready condition, here is it's status: {controller_pod['status']}"
+            if not ready:
```
osherdp, line 2450: would it be beneficial to also output in case of success? (controller is in ready status)

osherdp (changes requested):

tools/add_triage_signature.py
eranco74: :+1:

eranco74: Done

osherdp (approved):

PR #334, opened 2023-01-16
Add slow image download signature

eranco74:
See example [here](https://issues.redhat.com/browse/AITRIAGE-4999?jql=project%20in%20(12325045%2C%2012327120)%20AND%20text%20~%20%22Assisted%20Controller%20is%20not%20running%20on%20the%20cluster%22%20AND%20cf%5B12317358%5D%20!%3D%20%20SIGNATURE_missing_pivot_ostree%20ORDER%20BY%20created%20DESC%2C%20resolution%20ASC%2C%20updated%20DESC%2C%20component%20ASC)

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -27,3 +27,4 @@ aliases:
     - osherdp
     - omertuc
     - eliorerz
+    - eranco74
--- a/tools/add_triage_signature.py
+++ b/tools/add_triage_signature.py
@@ -839,6 +839,53 @@ class InstallationDiskFIOSignature(Signature):
             self._update_triaging_ticket(report)
 
 
+class SlowImageDownload(ErrorSignature):
+    image_download_regex = re.compile(
+        r"Host (?P<hostname>.+?): New image status (?P<image>.+?). result:.+?; download rate: (?P<download_rate>.+?) MBps"
+    )
+    minimum_download_rate_mb = 10
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            **kwargs,
+            function_impact_label="slow_image_download",
+            comment_identifying_string="h1. Host slow download rate:",
+        )
+
+    @classmethod
+    def _list_image_download_info(cls, events):
+        def get_image_download_info(event):
+            match = cls.image_download_regex.match(event["message"])
+            if match:
+                return match.groupdict()
+
+        return [image_info for event in events if (image_info := get_image_download_info(event)) is not None]
+
+    def _process_ticket(self, url, issue_key):
+        md = get_metadata_json(url)
+
+        cluster = md["cluster"]
+
+        cluster_id = cluster["id"]
+        events = get_cluster_installation_events(url, cluster_id)
+        image_info_list = self._list_image_download_info(events)
+
+        abnormal_image_info = []
+        for image_info in image_info_list:
+            if float(image_info["download_rate"]) < self.minimum_download_rate_mb:
+                abnormal_image_info.append(image_info)
+
+        if len(abnormal_image_info) > 0:
+            report = dedent(
+                """
+                        Detected slow image download rate (MBps):
+                        """
+            )
+            report += self._generate_table_for_report(abnormal_image_info)
+            self._update_triaging_ticket(report)
+
+
 class StorageDetailSignature(Signature):
     def __init__(self, *args, **kwargs):
         super().__init__(
@@ -2436,6 +2483,7 @@ class MachineConfigDaemonErrorExtracting(ErrorSignature):
 ############################
 JIRA_SERVER = "https://issues.redhat.com"
 ALL_SIGNATURES = [
+    SlowImageDownload,
     AllInstallationAttemptsSignature,
     EventsInstallationAttempts,
     ApiInvalidCertificateSignature,


OWNERS_ALIASES
```diff
@@ -27,3 +27,4 @@ aliases:
     - osherdp
     - omertuc
     - eliorerz
+    - eranco74
```
vrutkovs, line 30: sneaky :)

tools/add_triage_signature.py
```diff
@@ -839,6 +839,56 @@ def _process_ticket(self, url, issue_key):
             self._update_triaging_ticket(report)
 
 
+class DownloadImage(ErrorSignature):
+    image_download_regex = re.compile(r"Host (.+?): New image status (.+?). result:.+?; download rate: (.+?) MBps")
```
vrutkovs, line 843: This is "download + store on disk" rate effectively, right?

omertuc: I don't think the target file system for image pulls during discovery is disk bound

omertuc: Might be wrong though

omertuc: And if I'm wrong we should probably change it to be RAM backed

vrutkovs: Right, we do `podman pull` which includes "write on disk" time, but since it all happens on LiveCD we're writing into tmpfs iiuc, not "store on target disk"

tools/add_triage_signature.py
```diff
@@ -839,6 +839,56 @@ def _process_ticket(self, url, issue_key):
...
+        abnormal_image_info = []
+        for image_info in image_info_list:
+            if float(image_info["download_rate"]) < self.minimum_download_rate_mb:
+                abnormal_image_info.append(image_info)
+
+        if len(abnormal_image_info) > 0:
+            report = dedent(
+                """
+                        Detected slow image download rate (MBps):
+                        """
```
osherdp, line 886: seems like weird indentation

tools/add_triage_signature.py
```diff
@@ -839,6 +839,56 @@ def _process_ticket(self, url, issue_key):
             self._update_triaging_ticket(report)
 
 
+class DownloadImage(ErrorSignature):
```
osherdp, line 842: class name is not descriptive enough

tools/add_triage_signature.py
```diff
@@ -839,6 +839,56 @@ def _process_ticket(self, url, issue_key):
...
+    def _list_image_download_info(cls, events):
+        def get_image_download_info(event):
+            matches = cls.image_download_regex.findall(event["message"])
+            if len(matches) == 0:
+                return None
+            match = matches[0]
+            if len(match) < 3:
+                return None
+
+            return OrderedDict(id=event["host_id"], hostname=match[0], image=match[1], download_rate=match[2])
```
osherdp, line 864: python's dictionaries are already ordered since Python 3.7
```suggestion
            return dict(id=event["host_id"], hostname=match[0], image=match[1], download_rate=match[2])
```

tools/add_triage_signature.py
osherdp, line 843: let's use named groups, it makes it much more readable to parse it afterwards:
```suggestion
    image_download_regex = re.compile(r"Host (?P<hostname>.+?): New image status (?P<image>.+?). result:.+?; download rate: (?P<download_rate>.+?) MBps")
```

tools/add_triage_signature.py
```diff
@@ -839,6 +839,56 @@ def _process_ticket(self, url, issue_key):
...
+
+    @classmethod
+    def _list_image_download_info(cls, events):
+        def get_image_download_info(event):
+            matches = cls.image_download_regex.findall(event["message"])
+            if len(matches) == 0:
+                return None
+            match = matches[0]
+            if len(match) < 3:
+                return None
```
osherdp, line 862: how can this be? I think it shouldn't return a match if it doesn't have all 3 groups

osherdp (changes requested):

tools/add_triage_signature.py
eranco74: That's what block wants

tools/add_triage_signature.py
```diff
@@ -839,6 +839,56 @@ def _process_ticket(self, url, issue_key):
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            **kwargs,
+            function_impact_label="slow_image_download",
+            comment_identifying_string="h1. Host slow download rate:",
+        )
+
+    @classmethod
+    def _list_image_download_info(cls, events):
```
omertuc, line 855: I think this information is available in backend metadata.json, no need to parse events

omertuc: (.hosts[].images_status)

eranco74: For some reason this info in the cluster metadata is partial, we show the download rate only for some of the images.
check http://assisted-logs-collector.usersys.redhat.com/files/2023-01-13_13-56-56_158262a6-acea-4b94-9a71-bb2403675007/metadata.json

omertuc: Can you fix that / open a ticket to look into it?

omertuc: It doesn't bother anyone because we have [a dummy threshold](https://github.com/openshift/assisted-service/blob/5cc9607ca5e8154e4683fde764136fbe99a2c149/internal/host/validator.go#L55) for the validation

We get it from the image download [step reply](https://github.com/openshift/assisted-service/blob/76aab2c5d84a1833f3a4eeb10315af9f229ca18d/internal/bminventory/inventory.go#L3339-L3343)

eranco74: Looks like it's [here](https://github.com/openshift/assisted-test-infra/blob/962eed41792c2aff99adc616edbf409872478de8/src/assisted_test_infra/download_logs/download_logs.py#L328), it doesn't seem like we are doing something special for images_status.

PR #335, opened 2023-01-19
Remove CI reporting code

osherdp:
Now that implemented in
https://github.com/openshift-assisted/prow-jobs-scraper we no longer need the daily report and its code.


adriengentil (approved):
:+1:

PR #338, opened 2023-01-26
MGMT-13401: store cluster logs as Jira attachments

osherdp:
We plan on moving from the logs-collector server to a more maintainable and secure solution, being Jira attachments.

This change first creates Jira tickets as placeholder for each installation attempt (it no existing ticket can be found), archive and split the files (because Jira attachments have a size limit of 50MB) and continue with the rest of the process.

We shouldn't have lots of cases where we split the files (only 0.02% of the logs are more than 50MB), and for now this does not change the further processing through logs-collector server. Later on we'll create a method for reading Jira attachments and editing the signatures based on those.

### DIFF
--- a/Jenkinsfile.download_logs
+++ b/Jenkinsfile.download_logs
@@ -34,6 +34,9 @@ pipeline {
                     sh "scripts/install_environment.sh install_skipper"
                     sh "make image_build"
                 }
+
+                sh "rm -rf assisted-installer-deployment"
+                sh "git clone ${ASSISTED_INSTALLER_DEPLOYMENT_REMOTE} --branch ${ASSISTED_INSTALLER_DEPLOYMENT_BRANCH}"
             }
         }
 
@@ -42,14 +45,15 @@ pipeline {
                 dir ('assisted-test-infra') {
                     sh "make download_cluster_logs"
                 }
+
+                dir ('assisted-installer-deployment') {
+                    sh "skipper run ./tools/triage/attach_logs_in_jira.py --source-dir ${LOGS_DEST} --jira-access-token ${JIRA_ACCESS_TOKEN} --verbose"
+                }
             }
         }
 
         stage('Create tickets') {
             steps {
-                sh "rm -rf assisted-installer-deployment"
-                sh "git clone ${ASSISTED_INSTALLER_DEPLOYMENT_REMOTE} --branch ${ASSISTED_INSTALLER_DEPLOYMENT_BRANCH}"
-
                 dir ('assisted-installer-deployment') {
                     sh "skipper run ./tools/triage/create_triage_tickets.py --jira-access-token ${JIRA_ACCESS_TOKEN} -v"
                 }
--- a/requirements.in
+++ b/requirements.in
@@ -17,3 +17,4 @@ retry==0.9.2
 pytest==7.1.1
 networkx==2.5.1
 insights-core>3.0.276
+pyfakefs==5.1.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,13 +6,13 @@
 #
 async-timeout==4.0.2
     # via redis
-attrs==21.4.0
+attrs==22.2.0
     # via pytest
-cachecontrol==0.12.11
+cachecontrol[filecache]==0.12.11
     # via insights-core
-certifi==2022.6.15
+certifi==2022.12.7
     # via requests
-cffi==1.15.0
+cffi==1.15.1
     # via
     #   cryptography
     #   pynacl
@@ -20,7 +20,7 @@ charset-normalizer==2.0.12
     # via requests
 colorlog==4.7.2
     # via -r requirements.in
-cryptography==37.0.2
+cryptography==39.0.0
     # via secretstorage
 decorator==4.4.2
     # via
@@ -31,17 +31,19 @@ defusedxml==0.7.1
     #   insights-core
     #   jira
 deprecated==1.2.13
-    # via
-    #   pygithub
-    #   redis
+    # via pygithub
 fuzzywuzzy==0.18.0
     # via -r requirements.in
-idna==3.3
+idna==3.4
     # via requests
-iniconfig==1.1.1
+importlib-metadata==6.0.0
+    # via keyring
+iniconfig==2.0.0
     # via pytest
-insights-core==3.0.280
+insights-core==3.1.4
     # via -r requirements.in
+jaraco-classes==3.2.3
+    # via keyring
 jeepney==0.8.0
     # via
     #   keyring
@@ -50,12 +52,16 @@ jinja2==3.1.2
     # via insights-core
 jira==3.1.1
     # via -r requirements.in
-keyring==23.6.0
+keyring==23.13.1
     # via jira
 lockfile==0.12.2
-    # via insights-core
-markupsafe==2.1.1
+    # via
+    #   cachecontrol
+    #   insights-core
+markupsafe==2.1.2
     # via jinja2
+more-itertools==9.0.0
+    # via jaraco-classes
 msgpack==1.0.4
     # via cachecontrol
 multi-key-dict==2.0.3
@@ -64,14 +70,13 @@ nestedarchive==0.2.4
     # via -r requirements.in
 networkx==2.5.1
     # via -r requirements.in
-oauthlib==3.2.1
+oauthlib==3.2.2
     # via requests-oauthlib
 packaging==21.0
     # via
     #   -r requirements.in
     #   pytest
-    #   redis
-pbr==5.9.0
+pbr==5.11.1
     # via python-jenkins
 pluggy==1.0.0
     # via pytest
@@ -81,9 +86,11 @@ py==1.11.0
     #   retry
 pycparser==2.21
     # via cffi
+pyfakefs==5.1.0
+    # via -r requirements.in
 pygithub==1.55
     # via -r requirements.in
-pyjwt==2.4.0
+pyjwt==2.6.0
     # via pygithub
 pynacl==1.5.0
     # via pygithub
@@ -103,7 +110,7 @@ pyyaml==6.0
     # via
     #   -r requirements.in
     #   insights-core
-redis==4.3.4
+redis==4.4.2
     # via insights-core
 requests==2.26.0
     # via
@@ -119,7 +126,7 @@ requests==2.26.0
     #   requests-toolbelt
 requests-oauthlib==1.3.1
     # via jira
-requests-toolbelt==0.9.1
+requests-toolbelt==0.10.1
     # via
     #   jira
     #   python-gitlab
@@ -127,7 +134,7 @@ retry==0.9.2
     # via -r requirements.in
 ruamel-yaml==0.17.16
     # via -r requirements.in
-secretstorage==3.3.2
+secretstorage==3.3.3
     # via keyring
 six==1.16.0
     # via
@@ -140,10 +147,12 @@ tomli==2.0.1
     # via pytest
 tqdm==4.59.0
     # via -r requirements.in
-urllib3==1.26.9
+urllib3==1.26.14
     # via requests
 wrapt==1.14.1
     # via deprecated
+zipp==3.12.0
+    # via importlib-metadata
 
 # The following packages are considered to be unsafe in a requirements file:
 # setuptools
--- a/skipper.yaml
+++ b/skipper.yaml
@@ -7,5 +7,6 @@ volumes:
   - $PWD/assisted-installer.yaml:/assisted-installer.yaml:rw
   - $HOME/.ssh/known_hosts:/root/.ssh/known_hosts:ro
   - ${XDG_RUNTIME_DIR}/containers/auth.json:/run/user/0/containers/auth.json:ro
+  - /var/ai-logs:/var/ai-logs # using this when downloading triage logs
 env:
   REGISTRY_AUTH_FILE: /run/user/0/containers/auth.json
--- a/tests/tools/triage/test_attach_logs_in_jira.py
+++ b/tests/tools/triage/test_attach_logs_in_jira.py
@@ -1,0 +1,203 @@
+import re
+from pathlib import Path
+from unittest.mock import Mock, call, patch
+
+import jira
+import pytest
+from pyfakefs.fake_filesystem import FakeFilesystem
+
+from tools.triage.attach_logs_in_jira import attach_logs_in_jira
+
+
+@pytest.fixture(autouse=True)
+def jira_client():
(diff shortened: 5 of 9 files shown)


tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,98 @@
+import argparse
+import logging
+import re
```
omertuc, line 3: Unused

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Create a Jira ticket for each installation attempt, " "and upload artifacts as Jira attachments."
+    )
+
+    parser.add_argument(
+        "--jira-access-token",
+        default=os.environ.get("JIRA_ACCESS_TOKEN"),
```
omertuc, line 88: `os` undefined

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,98 @@
...
+        required=False,
+        help="PAT (personal access token) for accessing Jira",
+    )
+    parser.add_argument("--source-dir", help="Local directory with logs for all clusters")
+
+    attach_logs_in_jira(source_dir=args.source_dir, jira_access_token=args.jira_access_token)
```
omertuc, line 94: `args` undefined

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,98 @@
...
+        description="Create a Jira ticket for each installation attempt, " "and upload artifacts as Jira attachments."
+    )
+
+    parser.add_argument(
+        "--jira-access-token",
+        default=os.environ.get("JIRA_ACCESS_TOKEN"),
+        required=False,
+        help="PAT (personal access token) for accessing Jira",
+    )
+    parser.add_argument("--source-dir", help="Local directory with logs for all clusters")
```
omertuc, line 92:
```suggestion
    parser.add_argument("--source-dir", help="Local directory with logs for all clusters")
    
    args = parser.parse_args()
```

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,98 @@
...
+        read_chunk = partial(whole_file.read, chunk)
+        for i, chunk in enumerate(iter(read_chunk, "")):
+            with open(f"{path}-part-{i}", "wb") as part:
+                part.write(chunk)
+
+
+def attach_logs_in_jira(source_dir: Path, jira_access_token: str):
+    jira_client = jira.JIRA(JIRA_SERVER, token_auth=jira_access_token, validate=True)
+
+    for cluster_logs in source_dir.iterdir():
```
omertuc, line 59: naming nit: `s/cluster_logs/cluster_logs_dir/g`

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,98 @@
...
+        issue = _get_or_create_jira_ticket(jira_client, JIRA_SUMMARY.format(cluster_logs))
+
+        for artifact in cluster_logs.iterdir():
+            if artifact.is_dir():
+                archived_file = _archive_dir(artifact)
+                shutil.rmtree(artifact)
```
omertuc, line 65: So this script not only creates tickets, but it turns all directories within cluster logs directories into archives, destroying the original directory in the process? Seems a bit unexpected / violent for this script to do this

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,98 @@
...
+
+    if len(matching_issues) == 1:
+        log.debug("Using pre-existing issue %s", matching_issues[0])
+        return matching_issues[0]
+
+    raise RuntimeError(f"Found more than one matching issue for {cluster_logs}: {' '.join(matching_issues)}")
+
+
+def _archive_dir(directory: Path) -> Path:
+    archived_file = Path(f"{directory}.tar.gz")
```
omertuc, line 39: nit: with_suffix can save you formatting and the `Path` cast

```suggestion
    archived_file = directory.with_suffix(".tar.gz")
```

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,98 @@
...
+import shutil
+import tarfile
+from pathlib import Path
+from functools import partial
+
+import jira
+
+from consts import JIRA_PROJECT, JIRA_SUMMARY, JIRA_SERVER
+
+JIRA_ATTACHMENT_MAX_SIZE = 50 * 2**20  # 50MB
```
omertuc, line 13:
```suggestion
JIRA_ATTACHMENT_MAX_SIZE = 50 * 2**20  # 50MiB
```

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,98 @@
...
+                archived_file = _archive_dir(artifact)
+                shutil.rmtree(artifact)
+                artifact = archived_file
+
+            if artifact.stat().st_size > JIRA_ATTACHMENT_MAX_SIZE:
+                _split_file(artifact, chunk=JIRA_ATTACHMENT_MAX_SIZE)
+                artifact.unlink()
+
+        for artifact in cluster_logs.iterdir():
+            if any(attachment.name == artifact.name for attachment in issue.fields.attachment):
```
omertuc, line 73: Possibly be more readable with `in`:

```suggestion
            if artifact.name in (attachment.name for attachment in issue.fields.attachment):
```

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,98 @@
...
+    return archived_file
+
+
+def _split_file(path: Path, chunk: int) -> None:
+    log.debug("Splitting %s to several files", path)
+    with path.open("rb") as whole_file:
+        read_chunk = partial(whole_file.read, chunk)
+        for i, chunk in enumerate(iter(read_chunk, "")):
+            with open(f"{path}-part-{i}", "wb") as part:
+                part.write(chunk)
```
omertuc, line 53: Consider simply running the standard binutil `split`

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,98 @@
...
+def _archive_dir(directory: Path) -> Path:
+    archived_file = Path(f"{directory}.tar.gz")
+    log.debug("Archiving directory %s to %s", directory, archived_file)
+    with tarfile.open(archived_file, "w:gz") as tar:
+        tar.add(directory, directory.name)
+
+    return archived_file
+
+
+def _split_file(path: Path, chunk: int) -> None:
```
omertuc, line 47: Consider simply running the standard binutil `split`

tools/attach_logs_in_jira.py
osherdp: I'm changing it to not remove any file in the process

tools/attach_logs_in_jira.py
osherdp: thought about it, but as long as Python does the job well enough I prefer staying in that realm

osherdp:
This job still needs to be activated, and I need to add some tests. So it's not yet ready

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,114 @@
...
+def _archive_dir(directory: Path) -> Path:
+    archived_file = directory.with_suffix(".tar.gz")
+    log.debug("Archiving directory %s to %s", directory, archived_file)
+    with tarfile.open(archived_file, "w:gz") as tar:
+        tar.add(directory, directory.name)
+
+    return archived_file
+
+
+def _split_file(path: Path, chunk: int) -> None:
```
omertuc, line 54: naming nit: `s/chunk/chunk_size/g`

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,114 @@
...
+    log.debug("Splitting %s to several files", path)
+    with path.open("rb") as whole_file:
+        read_chunk = partial(whole_file.read, chunk)
+        for i, chunk in enumerate(iter(read_chunk, b"")):
+            with open(f"{path}-part-{i}", "wb") as part:
```
omertuc, line 59: naming nit: `s/part/part_file/g`

osherdp:
/hold
still pending a run in jenkins

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,116 @@
...
+        for i, chunk in enumerate(iter(read_chunk, b"")):
+            with open(f"{path}-part-{i}", "wb") as part_file:
+                part_file.write(chunk)
+
+
+def attach_logs_in_jira(source_dir: Path, jira_access_token: str):
+    jira_client = jira.JIRA(JIRA_SERVER, token_auth=jira_access_token, validate=True)
+
+    for cluster_logs_dir in source_dir.iterdir():
+        issue = _get_or_create_jira_ticket(jira_client, JIRA_SUMMARY.format(failure_id=cluster_logs_dir.name))
```
adriengentil, line 69:
```suggestion
        issue = _get_or_create_jira_ticket(jira_client, JIRA_SUMMARY.format(failure_id=cluster_logs_dir.name))
```
in order to make the process resilient, it could be good to catch the possible exception thrown by `_get_or_create_jira_ticket`, log and process the next folder

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,116 @@
...
+
+        for artifact in cluster_logs_dir.iterdir():
+            if artifact.is_dir():
+                artifact = _archive_dir(artifact)
```
adriengentil, line 73: should we check if an archive already exists before compressing the folder? As I understand, we don't cleanup the archives (which is fine I think).

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,116 @@
...
+            if artifact.stat().st_size > JIRA_ATTACHMENT_MAX_SIZE:
+                _split_file(artifact, chunk_size=JIRA_ATTACHMENT_MAX_SIZE)
+
+        for artifact in cluster_logs_dir.iterdir():
+            if artifact.is_dir() or artifact.stat().st_size > JIRA_ATTACHMENT_MAX_SIZE:
+                log.debug("Skip %s as it had been replaced with other files", artifact)
+                continue
+
+            if artifact.name in (attachment.filename for attachment in issue.fields.attachment):
+                log.debug("File %s already exists at %s. Skipping upload", artifact, issue)
```
adriengentil, line 84: Do we have cases where cluster logs are updated and we would need to upload an updated archive?

adriengentil (commented):
Looks good! Just few questions

tools/attach_logs_in_jira.py
osherdp: Good question. Maybe it's especially relevant for the events
I think I'll add another check of the size to understand if we need to upload updated files (no way to get a hash value like md5 as far as I can see, so it's the closets we can do apart from downloading the files from jira and do the hash calculation)

adriengentil: or embed the checksum in archive's name, if size check is easy, it could be good enough

osherdp: to keep it simple (and also file names shorter) I'll do the size check
either way it should be a low risk with low possibility

tools/attach_logs_in_jira.py
osherdp: yes it should be optimized

tools/attach_logs_in_jira.py
osherdp: Implemented something for it. I think in another PR I'll upgrade to python 3.11 to make use of ``ExceptionGroup`` (instead of how I do it with current code)

osherdp: https://github.com/openshift-assisted/assisted-installer-deployment/pull/344

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,158 @@
...
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Create a Jira ticket for each installation attempt, and upload artifacts as Jira attachments."
+    )
+
+    parser.add_argument("--source-dir", help="Local directory with logs for all clusters")
+    parser.add_argument("-v", "--verbose", action="store_true", help="Output verbose logging")
+    parser.add_argument(
+        "--days-ago-threshold",
```
osherdp, line 131: not the best naming in the world, need to rethink about it

adriengentil: `--max-days` ?

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,158 @@
...
+        log.debug("Using pre-existing issue %s", matching_issues[0])
+        return matching_issues[0]
+
+    raise RuntimeError(
+        f"Found more than one matching issue for '{summary}': {', '.join(issue.key for issue in matching_issues)}"
+    )
+
+
+def _archive_dir(directory: Path) -> Path:
+    archived_file = directory.with_suffix(".tar.gz")
```
omertuc, line 53: FYI I just learned that `with_suffix` removes the original extension. Disgusting. so if the directory ends with something like `.d` (which some directories often do) it will be deleted

osherdp: thanks!
to be on the safe side I'll handle this case

tools/attach_logs_in_jira.py
osherdp: yeah much better
thanks!

tools/attach_logs_in_jira.py
```diff
@@ -0,0 +1,158 @@
...
+                part_file.write(chunk)
+
+
+@retry.retry(exceptions=jira.JIRAError, tries=3, logger=log)
+def _attach_logs_for_cluster(jira_client: jira.JIRA, cluster_logs_dir: Path):
+    issue = _get_or_create_jira_ticket(jira_client, JIRA_SUMMARY.format(failure_id=cluster_logs_dir.name))
+
+    for artifact in cluster_logs_dir.iterdir():
+        if artifact.is_dir():
+            if artifact.with_suffix(".tar.gz").exists():
```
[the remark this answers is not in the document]
osherdp: thanks!

osherdp:
Seems like my initial assumption about current code being able to work with previously created tickets is not entirely true
In jenkins we only apply signatures and edit summary for newly created tickets, which I think is a bit limiting

Therefore, I'm moving this to draft mode until handling prerequisites. Feel free to unsubscribe for now as I'll cc back all who reviewed it so far when it's ready and tested
/uncc @adriengentil @omertuc carbonin mkowalski nmagnezi

osherdp:
/hold
waiting for reviews

adriengentil (approved):
Looks good :+1:

adriengentil (approved):

PR #342, opened 2023-02-01
Remove SMART column from host storage signature

omertuc:
I don't recall a single time where it was interesting and it keeps
breaking when we change the SMART command flags on the agent, it's
nothing but noise.


eranco74:
/lgtm
Hopefully we will remove it from the host inventory as well

omertuc:
Sure https://github.com/openshift/assisted-installer-agent/pull/489

nmagnezi (approved):
/lgtm

PR #345, opened 2023-02-05
filter-out placeholder tickets in triage status

osherdp:
the triage daily status generated a list containing tickets that are only placeholders for logs (generated by the testing for https://github.com/openshift-assisted/assisted-installer-deployment/pull/338). It included tickets for very old clusters that just weren't handled before, because we only filter by ticket creation time and not cluster creation time.

This changes the filter to exclude those kind of tickets (as they can exist for a short period of time also for new failures). In addition, it changes the filtering to be configured from code instead of a pre-defined filter (that currently cannot be managed because the owner left the company).

### DIFF
--- a/tools/triage_status_report.py
+++ b/tools/triage_status_report.py
@@ -11,8 +11,18 @@ import consts
 import jira
 import requests
 
-FILTER_ID = 12380672
 MISSING_VALUE = "<MISSING>"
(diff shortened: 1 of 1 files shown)


osherdp:
/hold
testing

osherdp:
/unhold
works

nmagnezi (approved):
/lgtm

PR #350, opened 2023-02-19
MGMT-13107: Create signature to detect MTU mismatch betwen interface and cluster network

ori-amizur:
In case of multi node cluster that uses SDN network type, there might be a situation that the cluster network uses jumbo frame MTU, and there is a host that doesn't use jumbo frames.  In this case SDN fails to initialize and the installation fails.
Original triaging ticket: https://issues.redhat.com/browse/AITRIAGE-4847

/cc @filanov 
/cc @omertuc

### DIFF
--- a/tools/triage/add_triage_signature.py
+++ b/tools/triage/add_triage_signature.py
@@ -2087,6 +2087,33 @@ class HostsInterfacesSignature(Signature):
         return interfaces_details
 
 
+class NetworksMtuMismatch(ErrorSignature):
+    LOG_PATTERN = re.compile(
+        r"Failed to start sdn: interface MTU [(]([0-9]+)[)] is too small for specified overlay MTU [(]([0-9]+)[)]"
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            **kwargs,
+            comment_identifying_string="h1. Networks MTU Mismatch",
+            function_impact_label="mgmt13107",
+        )
+
+    def _process_ticket(self, url, issue_key):
+        try:
+            sdn_logs = get_triage_logs_tar(triage_url=url, cluster_id=get_metadata_json(url)["cluster"]["id"]).get(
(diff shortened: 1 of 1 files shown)


tools/add_triage_signature.py
```diff
@@ -1943,6 +1943,30 @@ def _get_interfaces(self, host: dict):
...
 
+class NetworksMtuMismatch(ErrorSignature):
+    LOG_PATTERN = re.compile(r'Failed to start sdn: interface MTU [(]([0-9]+)[)]'
+                             r' is too small for specified overlay MTU [(]([0-9]+)[)]')
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            **kwargs,
+            comment_identifying_string="h1. Networks MTU Mismatch",
```
omertuc, line 1954: `ErrorSignature`s should also set `function_impact_label`

tools/add_triage_signature.py
```diff
@@ -1943,6 +1943,30 @@ def _get_interfaces(self, host: dict):
...
+        try:
+            sdn_logs = get_triage_logs_tar(
+                triage_url=url, cluster_id=get_metadata_json(url)["cluster"]["id"]
+            ).get("controller_logs.tar.gz/must-gather.tar.gz/must-gather.local.*/*/namespaces/openshift-sdn/pods/sdn-*/sdn/sdn/logs/*.log")
+        except FileNotFoundError:
+            return
+        match = self.LOG_PATTERN.search(sdn_logs)
+        if match:
+            self._update_triaging_ticket("SDN failed to start: Overlay (cluster) network MTU {}"
+                                         " is bigger than the interface MTU {}".format(match[2], match[1]))
```
omertuc, line 1967: nit: can be simplified

```suggestion
        if match := self.LOG_PATTERN.search(sdn_logs):
            self._update_triaging_ticket(f"SDN failed to start: Overlay (cluster) network MTU {match[2]}"
                                         " is bigger than the interface MTU {match[1]}")
```

ori-amizur: OK

osherdp (approved):

PR #351, opened 2023-02-21
MGMT-13748: fix git tag on stable promotions

osherdp:
In
https://github.com/openshift-assisted/assisted-installer-deployment/pull/336 I removed the git tagging of assisted-installer-deployment. It had happened because tagging the repo seems like an action that can happen from outside of the script.

Apparently, it's also used in QE process for approving snapshots of latest changes. We need to make sure we can do this tagging and that git credentials are passing as intended even when running on top of a container.

This change brings back the old code, in addition to providing credentials via ``GIT_ASKPASS``. This method should be more suitable to usage via containers (in oppose to SSH agents for example).

### DIFF
--- a/tools/assisted_installer_stable_promotion.py
+++ b/tools/assisted_installer_stable_promotion.py
@@ -1,12 +1,18 @@
 #!/usr/bin/env python3
 import argparse
+import contextlib
 import logging
+import tempfile
 import os
+import subprocess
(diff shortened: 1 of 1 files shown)


osherdp:
Seems to work for me locally, holding to see how QE wants to test it
/hold

adriengentil (approved):

PR #352, opened 2023-02-21
MGMT-13442 add generic issue filtering for triage

paul-maidment:
The aim of this PR is twofold:
* reorganize code so it's much easier to
  * run dryrun mode
  * filter issues by custom criteria
* add filtering by labels
* tests added


triage_resolving_filters.json
```diff
@@ -1,5 +1,12 @@
 {
-  "master_failed_to_pull_ignition_signature" : {
-    "MGMT-11089": "Master nodes failed to pull ignition"
+  "signatures": {
+    "master_failed_to_pull_ignition_signature" : {
+      "MGMT-11089": "Master nodes failed to pull ignition"
```
rccrdpccl, line 4: AFAIU this message will be looked for in the comments for closing. Is this sitll necessary? Should we get instead the message from the signature class, as [we already have that](https://github.com/openshift-assisted/assisted-installer-deployment/blob/master/tools/triage/add_triage_signature.py#L330)?

Makefile
```diff
@@ -41,7 +41,7 @@ flake8:
 	flake8 .
 
 unit-test:
-	pytest tools/
+	pytest --disable-warnings tests/
```
rccrdpccl, line 44: There a few warnings of deprecation of some libraries, we can get it resolved after closing this

dev-requirements.txt
```diff
@@ -12,7 +12,7 @@ autopep8==1.6.0
     # via -r dev-requirements.in
 backcall==0.2.0
     # via ipython
-black==22.6.0
+black==23.1.0
```
rccrdpccl, line 15: Needed for `match` statement

tools/jira_client/jira_dryrun_client.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def create_issue_link(self, link_name, issue, root_issue):
+        print(
+            f"Linked issue key={issue.key} as {link_name} root issue key={root_issue.key}\n",
+            file=self._dry_run_stream,
+            flush=True,
+        )
+
+    def transition_issue(self, issue, target_transition_id):
+        transition_names = {CLOSED_TRANSITION_ID: "Closed"}
```
osherdp, line 23: per above comment, it should be possible to just use status names instead of IDs, so this can be simplified

tools/jira_client/jira_api.py
```diff
@@ -0,0 +1,34 @@
...
+from .consts import CLOSED_TRANSITION_ID
+
+
+class JiraAPI:
+    def __init__(self, jira_client, logger) -> None:
+        self._jira_client = jira_client
+        self._logger = logger
+
+    def link_issue_to_root_issue(self, issue, root_issue):
+        self._logger.info("Linking issue key=%s to root issue key=%s", issue.key, root_issue.key)
```
osherdp, line 13:
```suggestion
        self._logger.info("Linking %s as related to %s", issue.key, root_issue.key)
```

rccrdpccl:
@osherdp can we test this in Jenkins before merging?

tools/jira_client/jira_api.py
```diff
@@ -0,0 +1,34 @@
...
+        self._jira_client.create_issue_link("relates to", issue.key, root_issue.key)
+
+    def close_issue(self, issue):
+        self._logger.info("Was requested to close issue %s ignoring for now", issue)
```
osherdp, line 17: sentence a bit weird, maybe just
```suggestion
        self._logger.info("Closing issue %s", issue)
```

tools/jira_client/jira_api.py
```diff
@@ -0,0 +1,34 @@
...
+        self._jira_client.transition_issue(issue.key, CLOSED_TRANSITION_ID)
+
+    def link_and_close_issue(self, issue, linked_issue=None):
+        if linked_issue:
+            self.link_issue_to_root_issue(issue, linked_issue)
+        self.close_issue(issue)
+
+    @retry.retry(exceptions=JIRAError, tries=3, delay=2)  # being resilient to 401 statuses
+    def get_issue_comments(self, issue):
+        if issue is None:
```
osherdp, line 27: why should it be none?

tools/jira_client/jira_api.py
```diff
@@ -0,0 +1,34 @@
...
+            return None
+        try:
+            return self._jira_client.comments(issue)
+        except JIRAError as e:
+            if "Issue Does Not Exist" not in str(e):
+                raise
+        return None
```
osherdp, line 34: I think that ultimately after 3 retries we should raise the error if we're consistently getting one
Otherwise who knows what potential issues we might miss

tools/triage/filters.py
```diff
@@ -0,0 +1,178 @@
...
+    def close_issues_from_json(jira_client, issues, json):
+        filters = Filters.from_json(jira_client, json)
+        jira_api = JiraAPI(jira_client, logger)
+        Filters.close_issues_by_filters(jira_api, issues, filters)
+
+    @staticmethod
+    def close_issues_from_args(jira_client, issues, args):
+        filters = Filters.from_args(jira_client, args)
+        jira_api = JiraAPI(jira_client, logger)
+        Filters.close_issues_by_filters(jira_api, issues, filters)
```
osherdp, line 178: I prefer the top level function to create the relevant filters (either from args or from json or from any other source) and provide them to a function of ``Filters.close_issues_by_filters`` to actually do the closing

tools/triage/filters.py
```diff
@@ -0,0 +1,178 @@
...
+class Filter:
+    def __init__(self, identifier, root_issue, message):
+        self.identifier = identifier
+        self.message = message
+        self.root_issue = root_issue
+
+    def matches(self, issue):
+        if issue.fields.status.name in ("Closed", "Done", "Obsolete"):
+            return False
+        return True
```
osherdp, line 28:
```suggestion
        return issue.fields.status.name not in ("Closed", "Done", "Obsolete")
```

osherdp (changes requested):
partial review

osherdp:
> @osherdp can we test this in Jenkins before merging?

yes, it's possible 
I'll send you the details in PM

tools/jira_client/jira_client_factory.py
```diff
@@ -0,0 +1,12 @@
...
+from .jira_dryrun_client import JiraDryRunClient
+
+
+class JiraClientFactory:
+    @staticmethod
+    def create(jira_access_token, dry_run_stream=None, validate=True):
+        client = jira.JIRA(server=consts.JIRA_SERVER, token_auth=jira_access_token, validate=validate)
+        if dry_run_stream is not None:
+            client = JiraDryRunClient(client, dry_run_stream)
+        return client
```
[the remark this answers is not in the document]
paul-maidment: We need to pass in a client into the dryrun client, this is why we don't just return immediately.

tools/jira_client/jira_api.py
paul-maidment: We kept the existing logic exactly as it was to avoid issues, we can change it if desired?

tools/jira_client/jira_api.py
paul-maidment: We kept the existing logic exactly as it was to avoid issues, we can change it if desired?

osherdp: now that it's ``jira_api`` it becomes more desirable to have the basics right
it makes this code more reusable than it was previously, so we wouldn't want people to rely on something that hides defects in it

tools/jira_client/jira_dryrun_client.py
```diff
@@ -0,0 +1,31 @@
...
+    def transition_issue(self, issue, target_transition_id):
+        transition_names = {CLOSED_STATUS: "Closed"}
+        transition_name = "Unknown"
+        if target_transition_id in transition_names.keys():
+            transition_name = transition_names[target_transition_id]
+        print(
+            f"Transitioned issue key={issue.issue.key} to {transition_name}({target_transition_id})",
+            file=self._dry_run_stream,
+            flush=True,
+        )
```
osherdp, line 31: why masking the status? I think it can just be:
```suggestion
        print(
            f"Transitioned issue key={issue.issue.key} to {target_transition_id},
            file=self._dry_run_stream,
            flush=True,
        )
```

tools/triage/add_triage_signature.py
```diff
@@ -18,6 +18,7 @@
 from collections import OrderedDict, defaultdict, Counter
 from datetime import datetime
 from textwrap import dedent
+from tools.jira_client import JiraClientFactory
```
osherdp, line 21: the convention we (mostly) follow for imports is splitting to:
1. built-ins
2. third-party libraries
3. internal code

So this should be right beside
```
from tools.triage.common import get_cluster_logs_base_url
```

(this applies here and elsewhere)

tools/triage/filters.py
```diff
@@ -0,0 +1,167 @@
...
+        if root_issue:
+            root_issue_key = root_issue.key
+        self.signature_instance = self._signature_class_name(jira_client, root_issue_key)
+        self.jira_api = JiraAPI(jira_client, logger)
+
+    def _get_signature_class_name(self, signature_class_alias_name):
+        signature_class_alias_name = signature_class_alias_name.lower().replace("-", "").replace("_", "")
+        signature_class_name = SIGNATURE_BY_TYPE.get(signature_class_alias_name)
+        assert signature_class_name is not None
+        return signature_class_name
```
osherdp, line 69: pretty much does the same
```suggestion
        return SIGNATURE_BY_TYPE[signature_class_alias_name]
```

tools/triage/filters.py
```diff
@@ -0,0 +1,167 @@
...
+    def create_filter(jira_client, filter_type, identifier, root_issue_key, message):
+        root_issue = jira_client.issue(root_issue_key)
+
+        match filter_type.lower():
+            case "labels":
+                new_filter = LabelFilter(identifier, root_issue, message)
+            case "signatures":
+                new_filter = SignatureFilter(jira_client, identifier, root_issue, message)
+            case other:
+                raise Exception(f"unknown filter type {other}")
```
osherdp, line 103:
```suggestion
                raise ValueError(f"unknown filter type {other}")
```

tools/triage/filters.py
```diff
@@ -0,0 +1,167 @@
...
+        return new_filter
+
+    @staticmethod
+    def from_json_filepath(jira_client, json_filepath):
+        with open(json_filepath) as fp:
+            filters_json = json.load(fp)
+            return Filters.from_json(jira_client, filters_json)
+        return None
```
osherdp, line 111:
```suggestion
```
delete. this should be dead-code

tests/tools/jira/test_jira_api.py
```diff
@@ -0,0 +1,82 @@
+import pytest
+
+from unittest.mock import Mock
+from tools.jira_client import JiraAPI, CLOSED_STATUS
+
+
+def get_mock_issue(key="foobar"):
+    issue = Mock()
+    issue.key = key
+    return issue
```
osherdp, line 10:
```suggestion
    return Mock(key=key)
```
maybe you don't even need this function as it's much shorter with its body

tests/tools/jira/test_jira_api.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def test_should_call_underlying_jira_client_to_link_issue_to_root_issue():
+    client = Mock()
+    logger = Mock()
+    client.create_issue_link = Mock()
```
osherdp, line 16:
```suggestion
```
all methods of a mock should automatically generate a mock object

tests/tools/jira/test_jira_api.py
```diff
@@ -0,0 +1,82 @@
...
+    jira_api = JiraAPI(jira_client=client, logger=logger)
+    comments = jira_api.get_issue_comments(issue=issue)
+    client.comments.assert_called_once_with(issue)
+    assert comments == mock_comments
+
+
+def test_get_comments_should_raise_exception_for_errors():
+    client = Mock()
+    logger = Mock()
+    jira_api = JiraAPI(jira_client=client, logger=logger)
```
osherdp, line 49: I see it's reoccurring in the same way, so you can do:
```python
from pytest import fixture

@pytest.fixture()
def jira_api():
    return JiraAPI(jira_client=Mock(), logger=Mock())

def test_should_call_underlying_jira_client_to_link_issue_to_root_issue(jira_api):
    issue = get_mock_issue("mock_issue")
    root_issue = get_mock_issue("root_issue")
    jira_api.link_issue_to_root_issue(issue=issue, root_issue=root_issue)
    jira_api._jira_client.create_issue_link.assert_called_once_with("relates to", "mock_issue", "root_issue")
```

tests/tools/jira/test_jira_client_factory.py
```diff
@@ -0,0 +1,19 @@
...
+from tools.jira_client import (
+   JiraClientFactory,
+   JiraDryRunClient
+)
+
+
+def test_should_create_real_jira_client_when_output_stream_is_none():
+    # This test does not need to validate credentials as we only care about the type of jira_client.
+    jira_client = JiraClientFactory.create("foobar", validate=False)
+    assert type(jira_client) is jira.client.JIRA
```
osherdp, line 12:
```suggestion
    assert isinstance(jira_client, jira.client.JIRA)
```

tests/tools/jira/test_jira_dryrun_client.py
```diff
@@ -0,0 +1,51 @@
+import os
+import pytest
+from unittest.mock import Mock
+from tools.jira_client.jira_api import CLOSED_STATUS
+from tools.jira_client.jira_dryrun_client import JiraDryRunClient
+
+
+def get_devnull_stream():
+    return open(os.devnull, "w")
```
osherdp, line 9:
```suggestion
@pytest.fixture()
def devnull_stream():
    with open(os.devnull, "w") as devnull:
        yield devnull
```

then tests can just have it as an argument. e.g.:
```python
def test_something(devnull):
...
```

tests/tools/jira/test_jira_dryrun_client.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def get_mock_issue(key="foobar"):
+    issue = Mock()
+    issue.key = key
+    return issue
+
+
+class MockAttributeErrorClient:
```
osherdp, line 18:
```suggestion
class MockAttributeErrorJiraClient:
```

tests/tools/jira/test_jira_dryrun_client.py
```diff
@@ -0,0 +1,51 @@
...
+    def method_exists(self):
```
osherdp, line 19: it's rather unused, no? so why do we need it for the tests?

tests/tools/jira/test_jira_dryrun_client.py
```diff
@@ -0,0 +1,51 @@
...
+    mock_jira_client = MockAttributeErrorClient()
+    jira_dryrun_client = JiraDryRunClient(jira_client=mock_jira_client, dry_run_stream=get_devnull_stream())
+    with pytest.raises(AttributeError):
+        jira_dryrun_client.does_not_exist()
+
+
+def test_should_proxy_calls_to_real_jira_when_calling_undefined_methods():
+    mock_jira_client = Mock()
+    mock_jira_client.search_issues = Mock(return_value=[])
+    jira_dryrun_client = JiraDryRunClient(jira_client=mock_jira_client, dry_run_stream=get_devnull_stream())
```
osherdp, line 33: just a second, the dry-run client is able to proxy any call to jira client? how can we make sure it doesn't do write actions through those calls?

tools/triage/filters.py
```diff
@@ -0,0 +1,169 @@
...
+
+class LabelFilter(Filter):
+    def matches(self, issue):
+        if not super().matches(issue):
+            return False
+        try:
+            logger.debug(f"{issue.key} labels: {issue.fields.labels}")
+            for label in issue.fields.labels:
+                if label == self.identifier:
+                    return True
```
osherdp, line 48:
```suggestion
            if self.identifier in issue.fields.labels:
                return True
```

tests/tools/triage/test_filters.py
```diff
@@ -0,0 +1,230 @@
...
+    issue.fields = MockFieldsWithNoLabels()
+    return issue
+
+
+def test_create_label_filter():
+    root_issue_key = "MGMT-0000"
+    jira_client = Mock()
+    jira_client.issue = Mock(return_value=get_mock_issue(root_issue_key))
+    triage_filter = Filters.create_filter(jira_client, "labels", "FEATURE_some_label", root_issue_key, "my message")
+    assert type(triage_filter) == LabelFilter
```
osherdp, line 38: use ``isinstance`` here and elsewhere

tests/tools/triage/test_filters.py
```diff
@@ -0,0 +1,230 @@
...
+    assert triage_filter.root_issue.key == root_issue_key
+    jira_client.issue.assert_called_with(root_issue_key)
+
+
+def test_create_invalid_type_filter():
+    root_issue_key = "MGMT-0000"
+    jira_client = Mock()
+    jira_client.issue = Mock()
+    with pytest.raises(Exception) as e_info:
+        _ = Filters.create_filter(jira_client, "invalid-type", "whatever", root_issue_key, "my message")
```
osherdp, line 58:
```suggestion
        Filters.create_filter(jira_client, "invalid-type", "whatever", root_issue_key, "my message")
```

tests/tools/triage/test_filters.py
```diff
@@ -0,0 +1,230 @@
...
+    assert type(triage_filter) == SignatureFilter
+    assert triage_filter.root_issue.key == root_issue_key
+    jira_client.issue.assert_called_with(root_issue_key)
+
+
+def test_create_invalid_type_filter():
+    root_issue_key = "MGMT-0000"
+    jira_client = Mock()
+    jira_client.issue = Mock()
+    with pytest.raises(Exception) as e_info:
```
osherdp, line 57:
```suggestion
    with pytest.raises(Exception, match="unknown filter type invalid-type"):
```

osherdp (changes requested):

tools/jira_client/jira_dryrun_client.py
rccrdpccl: Yeah, you are right, leftover from status_id and blindly replaced :sweat:

tests/tools/jira/test_jira_dryrun_client.py
rccrdpccl: It's the humans responsibility to add all the write methods to the JiraDryRun (just as we hide the write call behind an `if not dry run`, but automatically all read methods will work without the need to add them as boilerplate.

The alternative I see is to add _all_ the needed methods (read and write), and proxy the read ones manually.

The risk is that users will think this is a magic class and just using without adding write methods that might be used.

WDYT?

tools/triage/add_triage_signature.py
rccrdpccl: One day we'll have this file to pass pylint :sweat_smile:

rccrdpccl:
/hold
holding it because we need to revert a small change that I am using for testing in Jenkins

tests/tools/jira/test_jira_dryrun_client.py
osherdp: Yeah, I guess it's not that easy to change
Ultimately we can define an interface with ``abc`` to make sure all required methods are implemented. but I get how it's too much code
I also thought about a read-only jira tokens, but it requires us to create a new user because there's no read-only scope we can switch to

All in all, I think it's fine as-is

rccrdpccl: Got it, good point about the read token, we can keep it in mind for future iterations as I am aware this can potentially be dangerous.

rccrdpccl:
@osherdp I believe all comments are addressed now, we are ready when you are :)

tests/tools/jira/test_jira_dryrun_client.py
```diff
@@ -0,0 +1,52 @@
...
+from tools.jira_client.jira_dryrun_client import JiraDryRunClient
+
+
+@pytest.fixture()
+def devnull_stream():
+    with open(os.devnull, "w") as devnull:
+        yield devnull
+
+
+def get_mock_issue(key="foobar"):
```
osherdp, line 14: apply same suggestion as in a different case (``Mock(key=key)``)
and let's just get rid of those ``get_mock_issue`` functions. they are redundant and can be replaced with the body of the function

osherdp: this applies here and elsewhere, unless the function is more complex like in ``get_mock_issue``

tests/tools/triage/test_filters.py
```diff
@@ -0,0 +1,200 @@
...
+    triage_filter = Filters.create_filter(jira_client, "signatures", "master_failed_to_pull_ignition_signature", root_issue_key, "my message")
+    assert isinstance(triage_filter, SignatureFilter)
+    assert triage_filter.root_issue.key == root_issue_key
+    jira_client.issue.assert_called_with(root_issue_key)
+
+
+def test_create_invalid_type_filter():
+    root_issue_key = "MGMT-0000"
+    jira_client = Mock()
+    with pytest.raises(Exception, match="unknown filter type invalid-type"):
```
osherdp, line 44:
```suggestion
    with pytest.raises(ValueError, match="unknown filter type invalid-type"):
```

tests/tools/triage/test_filters.py
```diff
@@ -0,0 +1,200 @@
...
+    issues.append(get_mock_issue("ISSUE-4", "Open", ["some_label", "example_label"]))
+
+    filters = []
+    filters.append(
+        LabelFilter(
+            identifier="example_label",
+            root_issue=None,
+            message="this is another example message",
+        )
+    )
```
osherdp, line 119:
```suggestion
    filters = [
        LabelFilter(
            identifier="example_label",
            root_issue=None,
            message="this is another example message",
        )
    ]
```

osherdp: same goes for ``issues``. less mutations is better

tools/triage/add_triage_signature.py
osherdp: lot's of places still have this. can you have another look?

osherdp (changes requested):

tools/triage/add_triage_signature.py
rccrdpccl: :thinking:  it might be me but I don't see any out of order? Are you referring to this, right?

tests/tools/jira/test_jira_api.py
```diff
@@ -0,0 +1,71 @@
+import pytest
+
+from unittest.mock import Mock
```
osherdp, line 3: imports out of order

tests/tools/jira/test_jira_client_factory.py
```diff
@@ -0,0 +1,19 @@
+import sys
+import jira
+from tools.jira_client import (
+   JiraClientFactory,
+   JiraDryRunClient
+)
```
osherdp, line 6: should be:
```suggestion
import sys

import jira

from tools.jira_client import (
   JiraClientFactory,
   JiraDryRunClient
)
```

tests/tools/jira/test_jira_dryrun_client.py
```diff
@@ -0,0 +1,53 @@
+import os
+import pytest
+from unittest.mock import Mock
```
osherdp, line 3:
```suggestion
import os
from unittest.mock import Mock

import pytest
```

tests/tools/triage/test_filters.py
```diff
@@ -0,0 +1,197 @@
+import pytest
+import os
+
+from unittest.mock import Mock, call
```
osherdp, line 4: here as well, should be:
```suggestion
import os
from unittest.mock import Mock, call

import pytest
```

tools/jira_client/jira_client_factory.py
```diff
@@ -0,0 +1,13 @@
+# pylint: disable=too-few-public-methods
+import jira
+from tools import consts
+from .jira_dryrun_client import JiraDryRunClient
```
osherdp, line 4:
```suggestion
import jira

from tools import consts
from .jira_dryrun_client import JiraDryRunClient
```

tools/triage_status_report.py
```diff
@@ -4,12 +4,11 @@
 import json
 import os
 import sys
+import requests
 from typing import List
 from urllib import parse
+from tools.jira_client import JiraClientFactory
```
osherdp, line 10:
```suggestion
import json
import os
import sys
from typing import List
from urllib import parse

import requests

from tools.jira_client import JiraClientFactory
```

tools/triage/add_triage_signature.py
osherdp: adding some comments for those places (not for all though)

osherdp (changes requested):

tools/triage/add_triage_signature.py
rccrdpccl: Oh, thanks for pointing out, I was focusing only on the order of our packages :facepalm: 

Added `isort` so this can be done by a machine - which is much better than me at this job :grin:

osherdp (approved):
Very nice. Thank you!

tools/triage/add_triage_signature.py
osherdp: Yes, that's a great idea
Thank you!

osherdp (approved):

PR #353, opened 2023-02-22
MGMT-13752: Add signatures for existing triage tickets as well

osherdp:
Currently, the code at assisted-installer-deployment adds signatures and edit the description & fields only for newly created tickets.

We want to do the same for existing tickets for 2 main reasons:
* If the original job that created a ticket couldn't finish adding all signatures for that ticket (e.g. logs-collector server got down temporarily, we had a bug in one of the signatures that has been fixed, etc.) we want to give it another try in the next jobs.
* The plan for moving to Jira attachments relies on us creating tickets with the attachments and processing them further afterwards.

In addition, this changes some other things:

* Moving all triage-related logic to a directory ``tools/triage``.
* Instead of loading all issues and then matching them with the failure IDs, just search for each one individually.
* Remove now redundant code for normalizing failure IDs.
* Created a "openshift versions" signature, instead of setting it at ticket creation.
* Invoke the description signature (named ``FailureDescription``) just like any other signature (instead of invocation at ticket creation).


tools/utils.py
```diff
@@ -10,15 +9,3 @@ def get_credentials_from_netrc(*, hostname, netrc_file=None):
         return None, None
 
     return credentials[0], credentials[2]
-
-
-def normalize_jira_title(s):
```
osherdp, line 15: I expect all of this to be irrelevant, now that all new issues are normalized (and we only look at last month's issues)

tools/triage/create_triage_tickets.py
```diff
@@ -0,0 +1,108 @@
...
+            logs_url,
+            issue.key,
+            only_specific_signatures=None,
+            dry_run_file=None,
+        )
+        close_custom_domain_user_ticket(jira_client, issue.key)
+        close_by_signature.run_using_json(args.filters_json, jira_client, [issue])
+
+    if not args.filters_json:
+        return
```
adriengentil, line 72: leftover?

osherdp: oops. right

adriengentil (approved):
Looks good!

osherdp:
Thanks! Starting the testing phase in jenkins

osherdp:
I might do more changes. I'll let you know once it's all playing nicely in jenkins

osherdp:
@adriengentil it should be ready for another review

adriengentil (approved):

osherdp:
/unhold
test in jenkins passed, so unholding

PR #355, opened 2023-02-23
Retry per ticket handling errors

osherdp:
Seems like if we're having a single failure during processing of the tickets, we're going over all tickets again (even those that we already processed successfully).

This changes it to do the retries per ticket handling, to reduce the time needed.

### DIFF
--- a/tools/triage/add_triage_signature.py
+++ b/tools/triage/add_triage_signature.py
@@ -27,6 +27,7 @@ import requests
 import tqdm
 from fuzzywuzzy import fuzz
 from tabulate import tabulate
+from retry import retry
 
 from tools import consts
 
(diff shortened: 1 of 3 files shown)


omertuc:
It's seems that you're just adding more retries here, without removing, was this the intention?

osherdp:
> It's seems that you're just adding a new type of retries here, without removing any old type of retry, was this the intention?

I originally also removed the retries from the outer-scope, thinking we don't really need them:
https://github.com/openshift-assisted/assisted-installer-deployment/blob/05cf338f747039fa3d0ca441538bc816aef8ec99/tools/triage/create_triage_tickets.py#L38

But then I noticed that we also have Jira operations there (creating and searching for tickets). So removing those retries may incur in more flakes

Now, I'm realizing it can still be problematic to have the outer scope retries and that we can just apply it more specifically to the ``get_or_create_triage_ticket`` function. So let me adapt the code to do that

osherdp:
/unhold
tested in jenkins and it looks good

adriengentil (approved):

PR #357, opened 2023-02-28
append labels instead of resetting

rccrdpccl:
Lables were being resets, so it was impossible to add fake labels for testing for example (or other signatures labels could've been wiped)

### DIFF
--- a/tools/triage/add_triage_signature.py
+++ b/tools/triage/add_triage_signature.py
@@ -691,9 +691,8 @@ class FailureDetails(Signature):
             # project
             cluster_md = extract_cluster_md_from_existing_labels(issue)
 
-        existing_labels = issue.fields.__dict__["labels"] or []
+        existing_labels = issue.fields.labels
 
-        logger.info(f"ISSUE {issue.key} existing labels: {existing_labels}")
         update_fields = {
             "labels": existing_labels,
         }
--- a/tools/triage/add_triage_signature.py
+++ b/tools/triage/add_triage_signature.py
@@ -660,7 +660,7 @@ class FailureDetails(Signature):
         ]
 
     def _process_ticket(self, url, issue_key):
-        def extract_cluster_md_from_existing_labels():
+        def extract_cluster_md_from_existing_labels(issue):
             label_prefix_to_property = {
                 "AI_CLUSTER_": "id",
                 "AI_DOMAIN_": "email_domain",
@@ -668,9 +668,8 @@ class FailureDetails(Signature):
             }
             r = re.compile(r"(AI_[^_]*_)(.*)")
 
-            i = self._jira_client.issue(issue_key)
             cluster_md = {}
-            for label in i.fields.labels:
+            for label in issue.fields.labels:
                 m = r.match(label)
                 if m is None or len(m.groups()) != 2:
                     continue
@@ -680,6 +679,7 @@ class FailureDetails(Signature):
                 cluster_md[label_prefix_to_property[prefix]] = value
             return cluster_md
 
+        issue = self._jira_client.issue(issue_key)
         cluster_md = []
         try:
             md = get_metadata_json(url)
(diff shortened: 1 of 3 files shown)


rccrdpccl:
tested on jenkins:
* add needed labels if they get removed
* does not remove manually added labels

tools/triage/add_triage_signature.py
```diff
@@ -689,10 +689,13 @@ def extract_cluster_md_from_existing_labels():
             # have in the 'labels' field of the ticket
             # this is mainly relevant for triage tickets that were moved from 'MGMT' project to 'AITRIAGE'
             # project
-            cluster_md = extract_cluster_md_from_existing_labels()
+            cluster_md = extract_cluster_md_from_existing_labels(issue)
 
+        existing_labels = issue.fields.__dict__["labels"] or []
+
+        logger.info(f"ISSUE {issue.key} existing labels: {existing_labels}")
```
osherdp, line 696:
```suggestion
        logger.debug(f"Issue {issue.key} existing labels: {existing_labels}")
```

tools/triage/add_triage_signature.py
```diff
@@ -689,10 +689,13 @@ def extract_cluster_md_from_existing_labels():
             # have in the 'labels' field of the ticket
             # this is mainly relevant for triage tickets that were moved from 'MGMT' project to 'AITRIAGE'
             # project
-            cluster_md = extract_cluster_md_from_existing_labels()
+            cluster_md = extract_cluster_md_from_existing_labels(issue)
 
+        existing_labels = issue.fields.__dict__["labels"] or []
```
osherdp, line 694: ``issue.fields.labels`` doesn't do the trick?
I tried ``jira_client.issue("MGMT-13748").fields.labels`` and it worked even though there are no labels to this ticket

osherdp (changes requested):

tools/triage/add_triage_signature.py
rccrdpccl: For some reason we have encountered issues that when empty it throws an AttributeError, so we are doing it in this way, however it should be equivalent:

```
try:
    existing_issues = issue.fields.labels
except AttributeError:
    existing_issues = []
```
@osherdp would you prefer this form?

osherdp: > For some reason we have encountered issues that when empty it throws an AttributeError

Do you have an example?
I remember that when doing ``jira_client.search_issues`` we need to [explicitly](https://github.com/openshift-assisted/assisted-installer-deployment/blob/ddf9d8c1a8d2e0c84a7e766ff3f612964cbb9839/tools/triage/common.py#L19) mention all fields that should be loadd (otherwise they are missing). For ``jira_client.issue(<issue_key>)`` it shouldn't be required iiuc

I prefer making sure all labels are loaded correctly, because otherwise we can have the problem described in this PR (overriding labels)

rccrdpccl: Unfortunately we didn't save the issue key that was giving this error, but we can look for it, there should be one in the last few hundreds.

IIRC in our case the labels were indeed empty, but the labels field was raising an error instead of returning empty list or something

rccrdpccl: AITRIAGE-6262 has no labels for example (entering in the exception in our current code)

rccrdpccl: Oh you right - nevermind, there is a problem, we'll look into it

osherdp: I think the problem is that on some calls to ``search_issues`` we're missing ``fields=["labels"]``. e.g.:
https://github.com/openshift-assisted/assisted-installer-deployment/blob/ddf9d8c1a8d2e0c84a7e766ff3f612964cbb9839/tools/triage/add_triage_signature.py#L1160

I also previously had this problem and didn't know why I have it. But at least the problem is easy to fix (maybe we can centralize it in one function that makes sure we have all required labels)

PR #359, opened 2023-03-02
MGMT-13667: Signature to detect user login to the cluster hosts.

paul-maidment:
During installation, if the user logs into one of the cluster hosts this might be something we need to know about. Whereas we acknowledge that this does not prove that the user made a breaking change, it is an indicator that we should look closer for such things.

So we are adding this signature to detect a manual login to any of the cluster hosts.

### DIFF
--- a/tools/triage/add_triage_signature.py
+++ b/tools/triage/add_triage_signature.py
@@ -2362,6 +2362,42 @@ class ControllerWarnings(Signature):
             self._update_triaging_ticket("Controller logs contain some warnings:\n" + f"{{code}}{warning_text}{{code}}")
 
 
+class UserHasLoggedIntoCluster(Signature):
+    """
+    This signature will detect user login to the cluster.
+    """
+
+    USER_LOGIN_PATTERN = re.compile(r"session opened for user .+ by")
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            **kwargs,
+            comment_identifying_string="h1. User has logged into cluster nodes during installation",
+        )
+
+    def _process_ticket(self, url, issue_key):
+        cluster = get_metadata_json(url)["cluster"]
+        cluster_id = cluster["id"]
+
+        triage_logs_tar = get_triage_logs_tar(triage_url=url, cluster_id=cluster_id)
+
+        report = ""
+        for host in cluster["hosts"]:
+                    host_id = host["id"]
+
+                    try:
+                        journal_logs = get_host_log_file(triage_logs_tar, host_id, "journal.logs")
+                    except FileNotFoundError:
+                        logger.info("Could not get host log file for %s", host_id)
+                        continue
+                    
+                    if self.USER_LOGIN_PATTERN.findall(journal_logs):
+                        report += dedent(f"""h2. Host {host_id}: found evidence of a user login during installation. This might indicate that some settings have been changed manually, moreover if these settings were incorrect they could contribute to failure of the installation.""")
+                
+        if len(report) != 0:
+            self._update_triaging_ticket(report)
+
 class MissingOSTreePivot(ErrorSignature):
     """
     This signature looks for missing pivot URL in rpm-ostree status of the control-plane hosts
@@ -2595,6 +2631,7 @@ ALL_SIGNATURES = [
     MissingOSTreePivot,
     MachineConfigDaemonErrorExtracting,
     ControllerFailedToStart,
+    UserHasLoggedIntoCluster,
 ]
 
 ############################
--- a/tools/triage/add_triage_signature.py
+++ b/tools/triage/add_triage_signature.py
@@ -2384,19 +2384,22 @@ class UserHasLoggedIntoCluster(Signature):
 
         report = ""
         for host in cluster["hosts"]:
-                    host_id = host["id"]
+            host_id = host["id"]
 
-                    try:
-                        journal_logs = get_host_log_file(triage_logs_tar, host_id, "journal.logs")
-                    except FileNotFoundError:
-                        logger.info("Could not get host log file for %s", host_id)
-                        continue
-                    
-                    if self.USER_LOGIN_PATTERN.findall(journal_logs):
-                        report += dedent(f"""h2. Host {host_id}: found evidence of a user login during installation. This might indicate that some settings have been changed manually, moreover if these settings were incorrect they could contribute to failure of the installation.""")
-                
+            try:
+                journal_logs = get_host_log_file(triage_logs_tar, host_id, "journal.logs")
+            except FileNotFoundError:
+                logger.info("Could not get host log file for %s", host_id)
+                continue
+
+            if self.USER_LOGIN_PATTERN.findall(journal_logs):
+                report += dedent(
+                    f"""h2. Host {host_id}: found evidence of a user login during installation. This might indicate that some settings have been changed manually, moreover if these settings were incorrect they could contribute to failure of the installation."""
+                )
+
         if len(report) != 0:
             self._update_triaging_ticket(report)
+
 
 class MissingOSTreePivot(ErrorSignature):
     """


tools/triage/add_triage_signature.py
```diff
@@ -2362,6 +2362,46 @@ def _process_ticket(self, url, issue_key):
...
+                    try:
+                        journal_logs = get_host_log_file(triage_logs_tar, host_id, "journal.logs")
+                    except FileNotFoundError:
+                        continue
+                    
+                    if self.USER_LOGIN_PATTERN.findall(journal_logs):
+                        report += dedent(
+                            f"""
+                        h2. Host {host_id}: found evidence of a user login during installation"""
+                        )
```
osherdp, line 2400: one less indentation level here

paul-maidment:
cc: @omertuc 
cc: @osherdp 
cc: @tsorya

osherdp (changes requested):

tools/triage/add_triage_signature.py
```diff
@@ -2362,6 +2362,46 @@ def _process_ticket(self, url, issue_key):
...
+        super().__init__(
+            *args,
+            **kwargs,
+            comment_identifying_string="h1. Customer has logged into cluster nodes during installation",
+        )
+
+    def _process_ticket(self, url, issue_key):
+        md = get_metadata_json(url)
+
+        cluster = md["cluster"]
```
omertuc, line 2382: nit

```suggestion
        cluster = get_metadata_json(url)["cluster"]
```

tools/triage/add_triage_signature.py
```diff
@@ -2362,6 +2362,46 @@ def _process_ticket(self, url, issue_key):
...
+
+                    try:
+                        journal_logs = get_host_log_file(triage_logs_tar, host_id, "journal.logs")
+                    except FileNotFoundError:
+                        continue
+                    
+                    if self.USER_LOGIN_PATTERN.findall(journal_logs):
+                        report += dedent(
+                            f"""
+                        h2. Host {host_id}: found evidence of a user login during installation"""
```
omertuc, line 2399:
```suggestion
                        h2. Host {host_id}: found evidence of a user login during installation. <insert explanation of why this matters>"""
```

tools/triage/add_triage_signature.py
```diff
@@ -2362,6 +2362,44 @@ def _process_ticket(self, url, issue_key):
             self._update_triaging_ticket("Controller logs contain some warnings:\n" + f"{{code}}{warning_text}{{code}}")
 
 
+class UserHasLoggedIntoCluster(Signature):
+    """
+    This signature will detect customer login to the cluster.
+    """
+
+    USER_LOGIN_PATTERN = re.compile(r"Accepted publickey for core from")
```
omertuc, line 2370: This will only detect SSH logins. Can we detect password-based logins as well? Users can [set a password](https://github.com/openshift/assisted-service/blob/master/docs/set-discovery-password.md) on discovery ISOs

tools/triage/add_triage_signature.py
```diff
@@ -2362,6 +2362,45 @@ def _process_ticket(self, url, issue_key):
             self._update_triaging_ticket("Controller logs contain some warnings:\n" + f"{{code}}{warning_text}{{code}}")
 
 
+class UserHasLoggedIntoCluster(Signature):
+    """
+    This signature will detect customer login to the cluster.
+    """
+
+    USER_LOGIN_PATTERN = re.compile(r"publickey for core from")
+    # USER_LOGIN_PATTERN = re.compile(r"the")
```
omertuc, line 2371: ?

tools/triage/add_triage_signature.py
paul-maidment: After looking into this, I can see that whenever a manual login takes place, the phrase 

"...session opened for core by...."

Will be present irrespective of the authentication method and it looks like the journal logs usually show sessions opened for the root user during installation.

For this reason I think this is a fairly reasonable heuristic

WDYT?

tools/triage/add_triage_signature.py
```diff
@@ -2362,6 +2362,46 @@ def _process_ticket(self, url, issue_key):
             self._update_triaging_ticket("Controller logs contain some warnings:\n" + f"{{code}}{warning_text}{{code}}")
 
 
+class CustomerHasLoggedIntoCluster(Signature):
```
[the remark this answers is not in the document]
paul-maidment: done, the customer is now a user, everywhere

tools/triage/add_triage_signature.py
paul-maidment: I have updated this text to include some reason as to why this would be of interest.

tools/triage/add_triage_signature.py
osherdp: I think it still applies. Indentation is too much here

tools/triage/add_triage_signature.py
```diff
@@ -2362,6 +2362,44 @@ def _process_ticket(self, url, issue_key):
...
+        triage_logs_tar = get_triage_logs_tar(triage_url=url, cluster_id=cluster_id)
+
+        report = ""
+        for host in cluster["hosts"]:
+                    host_id = host["id"]
+
+                    try:
+                        journal_logs = get_host_log_file(triage_logs_tar, host_id, "journal.logs")
+                    except FileNotFoundError:
+                        logger.info("Could not get host log file for " + host_id)
```
osherdp, line 2392:
```suggestion
                        logger.info("Could not get host log file for %s", host_id)
```

osherdp (changes requested):

tools/triage/add_triage_signature.py
osherdp: If it catches both methods, I think it's good
we can even change it to ``re.compile(r"session opened for user .+ by")`` to make it future-proof for username changes

paul-maidment: agreed, came here to write this (more or less)

paul-maidment: PR updated to use this regex

tools/triage/add_triage_signature.py
paul-maidment: Done again, hopefully I got it this time

osherdp: nope. lint still fails on it:
```
./build/lib/tools/triage/add_triage_signature.py:2387:21: E117 over-indented
```

so just make sure lint check passes before merging

tools/triage/add_triage_signature.py
```diff
@@ -2362,6 +2362,42 @@ def _process_ticket(self, url, issue_key):
...
+                    host_id = host["id"]
+
+                    try:
+                        journal_logs = get_host_log_file(triage_logs_tar, host_id, "journal.logs")
+                    except FileNotFoundError:
+                        logger.info("Could not get host log file for %s", host_id)
+                        continue
+                    
+                    if self.USER_LOGIN_PATTERN.findall(journal_logs):
+                        report += dedent(f"""h2. Host {host_id}: found evidence of a user login during installation. This might indicate that some settings have been changed manually, moreover if these settings were incorrect they could contribute to failure of the installation.""")
```
osherdp, line 2396:
```suggestion
                host_id = host["id"]

                try:
                    journal_logs = get_host_log_file(triage_logs_tar, host_id, "journal.logs")
                except FileNotFoundError:
                    logger.info("Could not get host log file for %s", host_id)
                    continue
                    
                if self.USER_LOGIN_PATTERN.findall(journal_logs):
                    report += dedent(f"""h2. Host {host_id}: found evidence of a user login during installation. This might indicate that some settings have been changed manually, moreover if these settings were incorrect they could contribute to failure of the installation.""")
```

osherdp (changes requested):

tools/triage/add_triage_signature.py
osherdp: suggested a code-change in another commit. you can also apply it from UI

paul-maidment: OK, finally passing the lint

osherdp (approved):

PR #360, opened 2023-03-02
retry filter instantation if jira GET issue fails

rccrdpccl:
This would retry if `jira.issue(key)` fails for some reason.

We definitely need a better strategy, this is just to make it resilient for now

### DIFF
--- a/tools/triage/filters.py
+++ b/tools/triage/filters.py
@@ -1,6 +1,9 @@
 import json
 import logging
 
+import retry
+from jira.exceptions import JIRAError
+
 from tools.jira_client import JiraAPI
 
 from .add_triage_signature import ALL_SIGNATURES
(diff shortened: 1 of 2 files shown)


osherdp (approved):

tools/triage/filters.py
```diff
@@ -86,6 +89,7 @@ def __eq__(self, other):
 
 
 class Filters:
+    @retry.retry(exceptions=JIRAError, tries=3, delay=2)
     @staticmethod
```
osherdp, line 93:
```suggestion
    @staticmethod
    @retry.retry(exceptions=JIRAError, tries=3, delay=2)
```
seems to fix the problem

osherdp (changes requested):

tools/triage/filters.py
rccrdpccl: :facepalm: 
Thanks for the suggestion, I was preparing for a much bigger change after I saw the error

osherdp: I'd like changing those functions to just regular methods, but I'll guess it's a bigger change

osherdp (approved):

PR #361, opened 2023-03-05
Reduce warnings emitted for currently acceptable access

osherdp:
Current code for accessing internal files in the logs archive, produce warnings logs about access to the old location, which is yet to change in production system.

Those warnings are confusing to people reading those logs (or trying to debug locally their contributions), as it seems something wrong is happening in the code or the environment.

This changes the logs to ``DEBUG`` level for all cases (either logs are found in the new location or the older one), so that we'll still have the required information to understand where we're pulling logs from.


nmagnezi (approved):
/lgtm

PR #362, opened 2023-03-06
pass issue key instead of issue object from jira dry run client to real jira client

rccrdpccl:
Fix by passing key instead mistakenly passed whole issue


rccrdpccl:
duplicate of https://github.com/openshift-assisted/assisted-installer-deployment/pull/365

PR #364, opened 2023-03-08
Split long triage status reports

osherdp:
Slack is automatically splitting long messages (when it's more than 4040 characters long). Affects of this are not taht good, because it causes a malformed markdown format.

This change splits the report to several messages (one message after another. Unfortunately it's not seem possible to do in one thread when using webhooks and the webhook's client), when the table has more than 7 rows.
If there are 7 rows or less, it just fits it in one single message (as it's doing right now).

The change also leverages use of ``slack_sdk`` library, which should verify enable a better handling of requests and responses.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -18,3 +18,4 @@ pytest==7.1.1
 networkx==2.5.1
 insights-core>3.0.276
 pyfakefs==5.1.0
+slack_sdk==3.19.2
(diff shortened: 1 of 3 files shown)


osherdp:
/hold
waiting for tests and reviews

adriengentil (approved):
LGTM :+1:
if it pollutes the channel to much, we could also reply in a thread (like we do with he CI report)

PR #366, opened 2023-03-14
MGMT-13301: signature for detecting overlap nameservers

slaviered:
Based on case https://issues.redhat.com/browse/AITRIAGE-4979
- adding a signature to detect where there is a nameserver that belongs to the cluster network.


tools/triage/add_triage_signature.py
```diff
@@ -1103,6 +1103,81 @@ def _process_ticket(self, url, issue_key):
...
+
+    def _check_ip_in_cidr(self, ip_str, cidr_str):
+        ip_address = ipaddress.ip_address(ip_str)
+        network = ipaddress.ip_network(cidr_str, strict=False)
+        return ip_address in network
+
+    def _get_nameservers(self, url, base_path):
+        ns = set()
+        for resolvconf in self._get_resolv_conf_files(url, base_path):
+            PATTERN = "^nameserver (.*)$"
```
osherdp, line 1148: let's put it as class constant (under ``class NameserverInClusterNetwork(ErrorSignature):``). in here it's just weird

tools/triage/add_triage_signature.py
```diff
@@ -1103,6 +1103,81 @@ def _process_ticket(self, url, issue_key):
...
+
+        return files
+
+    def _check_ip_in_cidr(self, ip_str, cidr_str):
+        ip_address = ipaddress.ip_address(ip_str)
+        network = ipaddress.ip_network(cidr_str, strict=False)
+        return ip_address in network
+
+    def _get_nameservers(self, url, base_path):
+        ns = set()
```
osherdp, line 1146:
```suggestion
        nameservers = set()
```

tools/triage/add_triage_signature.py
```diff
@@ -1103,6 +1103,81 @@ def _process_ticket(self, url, issue_key):
...
+
+    def _process_ticket_helper(self, path, url, issue_key):
+        md = get_metadata_json(url)
+        cidrs = [network["cidr"] for network in md["cluster"]["cluster_networks"]]
+
+        report = ""
+        for nameserver in self._get_nameservers(url, path):
+            for cidr in cidrs:
+                if self._check_ip_in_cidr(nameserver, cidr):
+                    report += f"nameserver {nameserver} is in cidr {cidr}\n"
```
osherdp, line 1164: The details are good, but in this case I think we need an opening line about it. Otherwise it's not clear for people what ``nameserver {nameserver} is in cidr {cidr}`` should mean
Maybe ``User configured nameserver IPs which are within the range of the configured cluster-networks. Read more at MGMT-13301``

tools/triage/add_triage_signature.py
```diff
@@ -1103,6 +1103,81 @@ def _process_ticket(self, url, issue_key):
...
+            **kwargs,
+            function_impact_label="bad_nameserver",
+            comment_identifying_string="h1. Nameserver in internal network",
+        )
+
+    def _get_resolv_conf_files(self, url, base_path):
+        files = []
+        try:
+            triage_logs_tar = get_triage_logs_tar(triage_url=url, cluster_id=get_metadata_json(url)["cluster"]["id"])
+        except FileNotFoundError:
```
osherdp, line 1119: iiuc you shouldn't catch here the exception
otherwise how can you fallback to the older format at line 1172

osherdp (changes requested):

tools/triage/add_triage_signature.py
slaviered: I thought that the old/new format are part of the base_path while here we do not have logs at all. this is why I want to separate the cases. WDYT?

osherdp: ack. then it's ok as-is
thanks!

tools/triage/add_triage_signature.py
slaviered: DONE

osherdp (approved):

PR #371, opened 2023-04-02
Fix duplicated Jira attachments

osherdp:
Due to the retry mechanism of the Jira client, sometimes we can get into transient errors from the server that makes the client re-send attachments (or do any kind of POST request).

This bypasses the idempotent mechanism implemented in our code, one example is: https://github.com/openshift-assisted/assisted-installer-deployment/blob/036ed3a555abd3dc33c69f74ca42674a6d526aae/tools/triage/attach_logs_in_jira.py#L64-L69

### DIFF
--- a/tools/jira_client/jira_client_factory.py
+++ b/tools/jira_client/jira_client_factory.py
@@ -9,7 +9,7 @@ from .jira_dryrun_client import JiraDryRunClient
 class JiraClientFactory:
     @staticmethod
     def create(jira_access_token, dry_run_stream=None, validate=True):
-        client = jira.JIRA(server=consts.JIRA_SERVER, token_auth=jira_access_token, validate=validate)
+        client = jira.JIRA(server=consts.JIRA_SERVER, token_auth=jira_access_token, validate=validate, max_retries=0)
         if dry_run_stream is not None:
             client = JiraDryRunClient(client, dry_run_stream)
         return client


danielerez:
/lgtm

Maybe worth removing the mechanism mentioned in the description later (as a cleanup).

tools/jira_client/jira_client_factory.py
```diff
@@ -9,7 +9,7 @@
 class JiraClientFactory:
     @staticmethod
     def create(jira_access_token, dry_run_stream=None, validate=True):
-        client = jira.JIRA(server=consts.JIRA_SERVER, token_auth=jira_access_token, validate=validate)
+        client = jira.JIRA(server=consts.JIRA_SERVER, token_auth=jira_access_token, validate=validate, max_retries=0)
```
danielerez, line 12: So will we have some other indiciation in case of a request failure now?

osherdp: We have retries in some other places, e.g.:
https://github.com/openshift-assisted/assisted-installer-deployment/blob/dff5517458ae0770ffc1ec53ee1e2668aa44cf25/tools/triage/add_triage_signature.py#L2951
https://github.com/openshift-assisted/assisted-installer-deployment/blob/036ed3a555abd3dc33c69f74ca42674a6d526aae/tools/triage/attach_logs_in_jira.py#L45
https://github.com/openshift-assisted/assisted-installer-deployment/blob/f591723c55470fb910ca523093c44ca5eb73d37d/tools/triage/common.py#L13

The reason we "prefer" those retries is that they at least preserve idempotence and don't created redundant objects

As for indications, we could have debug logs for each retry. It's a good idea, I'll create a PR for that as well

osherdp:
> /lgtm
> 
> Maybe worth removing the mechanism mentioned in the description later (as a cleanup).

You mean removing this code? Because it's required for not having duplicated objects (issues, attachments, comments, etc.)

danielerez:
> > /lgtm
> > Maybe worth removing the mechanism mentioned in the description later (as a cleanup).
> 
> You mean removing this code? Because it's required for not having duplicated objects (issues, attachments, comments, etc.)

Oh I assumed that avoiding retries would make that code redundant. But seems indeed needed if we retry in other places.

PR #373, opened 2023-04-05
MGMT-13543: Add signature to detect address change after reboot

ori-amizur:
Add signature to detect address change for after reboot.  The signature scans the inventory, and then tries to detect if a leased DHCP allocated address was different from address discovered in the discovery phase.

/cc @filanov

### DIFF
--- a/tools/triage/add_triage_signature.py
+++ b/tools/triage/add_triage_signature.py
@@ -4,6 +4,7 @@
 import abc
 import argparse
 import functools
+import gzip
 import ipaddress
 import itertools
 import json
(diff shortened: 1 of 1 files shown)


tools/triage/add_triage_signature.py
```diff
@@ -1323,6 +1324,78 @@ def _process_ticket(self, url, issue_key):
             self._update_triaging_ticket(report)
 
 
+class IpChangedAfterReboot(ErrorSignature):
```
osherdp, line 1327: is this problematic in any way that the IP changed after the reboot?
if so, in what way?

omertuc, line 1327: How is this signature different from [this](https://github.com/openshift/assisted-installer/blob/61e5f341f575a01c3fe6f8851224a389b62df261/src/common/common.go#L175) existing warning?

ori-amizur: This works after node joins.  The signature detects also cases if node didn't join.

ori-amizur: IP address change is unexpected.  The nodes are identified by their IPs. Certificates are generated based on the IPs discovered during discovery phase.  One of the things that might happen is that one node can configure one of the VIPs to be its local IP.
In the case I worked on, crio tried to bind one of the IPs but it couldn't because the configured IP returned to the discovered IP.  So crio failed to start.

osherdp: Thanks for the explanation!

osherdp (approved):
LGTM

PR #374, opened 2023-04-13
Adds ConfiguringHostAndMachineConfigServerError signature

carbonin:
This signature checks for when a host is stuck in "Configuring" when the cluster fails and when the MCS also has logs about failing to serve ignition.

This indicates the likely root cause of the install failure

https://issues.redhat.com/browse/AITRIAGE-6838 is an example issue

### DIFF
--- a/tools/triage/add_triage_signature.py
+++ b/tools/triage/add_triage_signature.py
@@ -2686,6 +2686,44 @@ class MachineConfigDaemonErrorExtracting(ErrorSignature):
             self._process_ticket_helper(url, old_logs_path)
 
 
+class ConfiguringHostAndMachineConfigServerError(ErrorSignature):
+    """
+    Looks for a host that timed out in "Configuring" in addition to errors in the MCS logs
+    """
+
+    mcs_error = re.compile(r"couldn't get config for req")
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            **kwargs,
+            function_impact_label="configuring_host_mcs_error",
+            comment_identifying_string="h1. machine-config-server failed to serve ignition",
+        )
+
+    def _process_ticket(self, url, issue_key):
+        cluster = get_metadata_json(url)["cluster"]
+
+        if not any(host["progress"]["current_stage"] == "Configuring" for host in cluster["hosts"]):
+            return
+
+        try:
+            mco_pods = get_triage_logs_tar(triage_url=url, cluster_id=cluster["id"]).get(
+                "controller_logs.tar.gz/must-gather.tar.gz/must-gather.local.*/*/namespaces/openshift-machine-config-operator/pods"
+            )
+        except FileNotFoundError:
+            logger.debug("failed to find mcs pods in must-gather")
+            return
+
+        for log in mco_pods.glob("machine-config-server-*/**/*.log"):
+            with open(log, "r") as f:
+                if self.mcs_error.search(f.read()):
+                    self._update_triaging_ticket(
+                        'machine-config-server logs seem to indicate a failure to serve ignition. This may be why hosts are stuck in "Configuring"'
+                    )
+                    break
+
+
 ############################
 # Common functionality
 ############################
@@ -2737,6 +2775,7 @@ ALL_SIGNATURES = [
     MachineConfigDaemonErrorExtracting,
     ControllerFailedToStart,
     UserHasLoggedIntoCluster,
+    ConfiguringHostAndMachineConfigServerError,
 ]
 
 ############################


carbonin:
In my dry run, I've only found 6 matching instances in the first ~3000 tickets, so I'll leave it up to you to decide if it's worth adding this.

cc @osherdp @omertuc

carbonin:
Final tally was 17/6513 issues had this problem.

tools/triage/add_triage_signature.py
```diff
@@ -2686,6 +2686,46 @@ def _process_ticket(self, url, issue_key):
             self._process_ticket_helper(url, old_logs_path)
 
 
+class ConfiguringHostAndMachineConfigServerError(ErrorSignature):
+    """
+    Looks for a host that timed out in "Configuring" in addition to errors in the MCS logs
+    """
+
+    mcs_error = re.compile(r"couldn't get config for req")
```
omertuc, line 2694: Can we try to look for any error rather than this particular one?

tools/triage/add_triage_signature.py
```diff
@@ -2686,6 +2686,46 @@ def _process_ticket(self, url, issue_key):
...
+        super().__init__(
+            *args,
+            **kwargs,
+            comment_identifying_string="h1. machine-config-server failed to serve ignition",
+        )
+
+    def _process_ticket(self, url, issue_key):
+        md = get_metadata_json(url)
+
+        cluster = md["cluster"]
```
omertuc, line 2706: nit

```suggestion
        cluster = get_metadata_json(url)["cluster"]
```

tools/triage/add_triage_signature.py
```diff
@@ -2686,6 +2686,46 @@ def _process_ticket(self, url, issue_key):
...
+
+        hosts_in_configuring = False
+        for host in cluster["hosts"]:
+            stage = host["progress"]["current_stage"]
+            logger.debug("found host with current stage {}".format(stage))
+            hosts_in_configuring = hosts_in_configuring or (stage == "Configuring")
+
+        if not hosts_in_configuring:
+            logger.debug("found no hosts in configuring")
+            return
```
omertuc, line 2716: can be simplified

```suggestion
        if not any(
            host["progress"]["current_stage"] == "Configuring" for host in cluster["hosts"]
        ):
            return
```

tools/triage/add_triage_signature.py
```diff
@@ -2686,6 +2686,46 @@ def _process_ticket(self, url, issue_key):
...
+    Looks for a host that timed out in "Configuring" in addition to errors in the MCS logs
+    """
+
+    mcs_error = re.compile(r"couldn't get config for req")
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            **kwargs,
+            comment_identifying_string="h1. machine-config-server failed to serve ignition",
```
omertuc, line 2700: `ErrorSignature`s should have function impact as well (see other `ErrorSignature`s for examples)

tools/triage/add_triage_signature.py
```diff
@@ -2686,6 +2686,46 @@ def _process_ticket(self, url, issue_key):
...
+            logger.debug("found host with current stage {}".format(stage))
+            hosts_in_configuring = hosts_in_configuring or (stage == "Configuring")
+
+        if not hosts_in_configuring:
+            logger.debug("found no hosts in configuring")
+            return
+
+        try:
+            mcs_logs = get_triage_logs_tar(triage_url=url, cluster_id=md["cluster"]["id"]).get(
+                "controller_logs.tar.gz/must-gather.tar.gz/must-gather.local.*/*/namespaces/openshift-machine-config-operator/pods/machine-config-server-*/machine-config-server/machine-config-server/logs/current.log"
```
carbonin, line 2720: Given https://github.com/omertuc/nestedarchive/issues/2 I think I need to change this.

IIUC this will only give me the output of the first matching file, right?
We probably need to iterate through each replica of the mcs and search the logs for each of them.

omertuc: Right. If you give it the path of a directory you get a temp dir mirroring the dir in the archive

tools/triage/add_triage_signature.py
carbonin: I'm not sure how I would create a reliable pattern for this without having more examples of what we'd want to match. What I have right now is this:

```
2023-04-11T14:06:35.864269135Z E0411 14:06:35.864247       1 api.go:128] couldn't get config for req: {master 0xc0007f4580}, error: could not fetch config , err: machineconfig.machineconfiguration.openshift.io "" not found
2023-04-11T14:06:40.865019524Z I0411 14:06:40.864987       1 api.go:109] Pool master requested by address:"192.168.156.13:39550" User-Agent:"Ignition/2.14.0" Accept-Header: "application/vnd.coreos.ignition+json;version=3.3.0, */*;q=0.1"
```

carbonin:
>Final tally was 17/6513 issues had this problem.

I don't think this is worth adding. Closing this.

PR #376, opened 2023-05-10
fix pip-freeze dependencies

rccrdpccl:
pip-freeze has pip-compile as dependency, and it was broken. Pip-tools 6.13.0 or greater was required leading to a loop (unable to freeze dependency because pip-freeze wouldn't work).

This PR fixes this issue.


Dockerfile.assisted-installer-deployment
```diff
@@ -7,6 +7,7 @@ RUN dnf install -y \
     make \
     skopeo \
     libffi-devel \
+    krb5-devel \
```
rccrdpccl, line 10: This was a requirement from pip-compile, not sure why though

osherdp: what do you see if it's not present?

rccrdpccl: > what do you see if it's not present?

Error when executing pip-compile (executed by make pip-freeze): `krb5-config: command not found`

rccrdpccl: I think it comes from pip-compile, but I was also trying to add another unrelated dependency. I'll make sure it comes from pip-compile

rccrdpccl: whoops, it was indeed coming from the other dependency. Mystery solved :sweat_smile:

osherdp (approved):

PR #386, opened 2023-06-18
No issue add podman and docker shortcuts

paul-maidment:


Makefile
```diff
@@ -6,6 +6,15 @@ clean:
 	rm -rf build dist *egg-info ./__pycache__
 	find -name *.pyc -delete
 
+#
+# Triage search tool
+#
+prepare_search:
+	$(container_command) build . -t ticket-search-container -f Dockerfile.assisted-installer-deployment
```
osherdp, line 13: seems to just build the assisted-installer-deployment, nothing specific to the ticket-search tool

Makefile
```diff
@@ -6,6 +6,15 @@ clean:
...
+
+ticket_search:
+	@$(container_command) run -v ${PWD}/data/triage-tools-tickets:/data/triage-tools-tickets -e JIRA_ACCESS_TOKEN=${JIRA_ACCESS_TOKEN} -it ticket-search-container:latest ticket_search -c $(c) -p $(p)
```
osherdp, line 16: can we use the fully-named argument names? ``-c`` and ``-p`` are a bit cryptic

osherdp: (as well as ``$(c)`` and ``$(p)`` )

osherdp (changes requested):

Makefile
paul-maidment: It pretty much does, yes, I just wanted to keep this simple for the user. 
The target is "prepare_search" how we make that happen can be hidden from the user?

Makefile
paul-maidment: Done

Makefile
paul-maidment: Another thing worth noting is that I don't want to build this target using skipper, the whole point of this PR is that we don't use skipper. The makefile does not have a target that just builds this, so I think this is OK

osherdp: Why "make prepare_search" is better than "make image"? Not sure I follow

paul-maidment: Fair point on the naming convention, I have updated now.

osherdp (approved):

PR #387, opened 2023-06-21
Match more possible controller log file names

carbonin:
Before this some controller logs were getting missed because they were named just `assisted-installer-controller.logs` without any `-`.

This meant that signatures like the one for UMN and LB-related operator failures were not being run.

cc @omertuc @paul-maidment @rccrdpccl


omertuc:
For the record the regression was due to https://github.com/openshift/assisted-installer/commit/7a9e909bc

PR #388, opened 2023-06-22
MGMT-13650: Change MissingOSTreePivot signature to OSTreeCommitMismatch

javipolo:
Check if ostree commit matches the one in the release_image for the openshift version to install

This cover the case when pivot is not performed as well as when pivot is not needed

### DIFF
--- a/Jenkinsfile.download_logs
+++ b/Jenkinsfile.download_logs
@@ -20,6 +20,7 @@ pipeline {
         OFFLINE_TOKEN = credentials('admin_offline_token')
         JIRA_ACCESS_TOKEN = credentials('assisted-installer-bot-jira-access-token')
         SLACK_TOKEN = credentials('slack-token')
+        PULL_SECRET_CONTENTS = credentials('prod-pull-secret-text')
     }
     options {
       timeout(time: 90, unit: 'MINUTES')
@@ -55,7 +56,7 @@ pipeline {
         stage('Create tickets') {
             steps {
                 dir ('assisted-installer-deployment') {
-                    sh "skipper run ./tools/triage/create_triage_tickets.py --jira-access-token ${JIRA_ACCESS_TOKEN} -v"
+                  sh 'skipper run -e PULL_SECRET_CONTENTS ./tools/triage/create_triage_tickets.py --jira-access-token ${JIRA_ACCESS_TOKEN} -v'
                 }
             }
         }
--- a/tools/triage/add_triage_signature.py
+++ b/tools/triage/add_triage_signature.py
@@ -282,6 +282,7 @@ class Signature(abc.ABC):
         dry_run_file=None,
         should_reevaluate=False,
         old_comment_string=None,
+        pull_secret_file=None,
     ):
         self._jira_client: jira.JIRA = jira_client
         self._identifing_string = comment_identifying_string
@@ -289,6 +290,7 @@ class Signature(abc.ABC):
         self.dry_run_file = dry_run_file
         self.should_reevaluate = should_reevaluate
         self.issue_key = issue_key
+        self.pull_secret_file = pull_secret_file
 
     def process_ticket(self, url, issue_key):
         try:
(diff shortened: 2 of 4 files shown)


tools/triage/add_triage_signature.py
```diff
@@ -2626,41 +2627,83 @@ def _process_ticket(self, url, issue_key):
...
         )
 
+    def get_os_content_image(self, release_image):
+        command = ["oc", "adm", "release", "info", release_image, "-ojson"]
+        if self.pull_secret:
+            command += ["--registry-config", self.pull_secret]
+        release_json = json.loads(subprocess.check_output(command))
+        tags = release_json["references"]["spec"]["tags"]
+        image_name = next(name for name in self.machine_os_content_names if name in [tag["name"] for tag in tags])
+        return next(tag["from"]["name"] for tag in tags if tag["name"] == image_name)
```
paul-maidment, line 2659: Maybe a note here to say that the first tag matching one of the machine_os_content_names and the first name within that tag is OK as this is what we are expecting?

The iterator (next) syntax is quite hard to understand at first, I spent some time playing around and I now get it, but to the casual eye, this code is quite hard to read.

With that said, I like the compact nature.

I see that you have not specified a default for next() so will encounter a StopIteration error if none of the tags match, I think it would be an idea to handle this gracefully somehow so that you can tell the user that you could not find any of the machine_os_content names in tags.

tools/triage/add_triage_signature.py
```diff
@@ -2626,41 +2627,83 @@ def _process_ticket(self, url, issue_key):
...
+        command = [
+            "skopeo",
+            "inspect",
+            "--config",
+            f"docker://{image}",
+        ]
+        if self.pull_secret:
+            command += ["--authfile", self.pull_secret]
+        labels = json.loads(subprocess.check_output(command))["config"]["Labels"]
+        return next(labels[label] for label in self.ostree_commit_labels if label in labels)
```
paul-maidment, line 2671: Again, maybe explain here that we are only interested in the first matching label from self.ostree_commit_labels, presumably because if this is in order or priority and mutually exclusive.

paul-maidment (approved):
I am actually giving this an approval but would suggest maybe adding some error messages for extreme edge cases as suggested in comments.

tools/triage/add_triage_signature.py
javipolo: Agree on both

- code explained in a comment
- and controlled with a try catch when calling the function (I dont know if it's the best way to handle this ....)

tools/triage/add_triage_signature.py
```diff
@@ -2626,41 +2627,83 @@ def _process_ticket(self, url, issue_key):
...
             node_ip = os.path.basename(node_dir)
             try:
-                # Fetch ostree status file for the node
-                ostree_status_file = triage_logs_tar.get(f"{path}/{node_ip}/rpm-ostree/status")
+                # Get rpm-ostree current deployment from status file
+                lines = triage_logs_tar.get(f"{path}/{node_ip}/rpm-ostree/status").splitlines()
+                ostree_status_line = re.sub(
+                    r"^. (ostree:\/\/|ostree-unverified-registry:|pivot:\/\/)?",
+                    "",
+                    next(lines[i + 1].strip() for i, line in enumerate(lines[:-1]) if line.strip() == "Deployments:"),
```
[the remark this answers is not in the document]
javipolo: Set a default that can be used as an error message to be displayed

tools/triage/add_triage_signature.py
javipolo: Done

Jenkinsfile.download_logs
```diff
@@ -55,7 +55,9 @@ pipeline {
         stage('Create tickets') {
             steps {
                 dir ('assisted-installer-deployment') {
-                    sh "skipper run ./tools/triage/create_triage_tickets.py --jira-access-token ${JIRA_ACCESS_TOKEN} -v"
+                    withCredentials([file(credentialsId: 'prod-pull-secret', variable: 'PULL_SECRET')]) {
+                      sh "skipper run ./tools/triage/create_triage_tickets.py --jira-access-token ${JIRA_ACCESS_TOKEN} --pull-secret ${PULL_SECRET} -v"
```
osherdp, line 59:
```suggestion
                      sh "skipper run ./tools/triage/create_triage_tickets.py --jira-access-token ${JIRA_ACCESS_TOKEN} -v"
```

we can probably remove this as it defaults to ``os.environ.get("PULL_SECRET")`` in the code, no?

tools/triage/create_triage_tickets.py
```diff
@@ -95,6 +95,12 @@ def main(args):
         required=False,
         help="PAT (personal access token) for accessing Jira",
     )
+    loginArgs.add_argument(
+        "--pull-secret",
+        default=os.environ.get("PULL_SECRET"),
```
osherdp, line 100: just to make it clearer from outside:
```suggestion
        "--pull-secret-file",
        default=os.environ.get("PULL_SECRET_FILE"),
```

tools/triage/add_triage_signature.py
```diff
@@ -2626,60 +2627,119 @@ def _process_ticket(self, url, issue_key):
...
+        release_json = json.loads(subprocess.check_output(command))
+        tags = release_json["references"]["spec"]["tags"]
+        try:
+            # Get the first tag present in the release image that matches the machine_os_content list
+            # Order is important, since machine-os-content tag might be present on release_images where the correct one is rhel-coreos for example
+            image_name = next(name for name in self.machine_os_content_names if name in [tag["name"] for tag in tags])
+            # And return the image for that tag
+            return next(tag["from"]["name"] for tag in tags if tag["name"] == image_name)
+        except StopIteration:
+            return None
```
osherdp, line 2664: maybe a regular ``for`` loop over ``tags`` will be clearer here, instead of usage of ``next`` and the stoppage at ``StopIteration``
any reason it won't work?

tools/triage/add_triage_signature.py
```diff
@@ -2626,60 +2627,119 @@ def _process_ticket(self, url, issue_key):
...
+        command = [
+            "skopeo",
+            "inspect",
+            "--config",
+            f"docker://{image}",
+        ]
+        if self.pull_secret:
+            command += ["--authfile", self.pull_secret]
+        labels = json.loads(subprocess.check_output(command))["config"]["Labels"]
+        # Return the first ocurrence that matches a label in self.ostree_commit_labels
```
osherdp, line 2676: nit
```suggestion
        # Return the first occurrence that matches a label in self.ostree_commit_labels
```

Jenkinsfile.download_logs
osherdp: if not it should be possible to just do ``export PULL_SECRET`` the line before. I think there's no reason it won't work

javipolo: In the end export was needed, and I preferred to explicitly add the command line argument

Also, the file needed to be copied in a directory that was accessible from within skipper run

Jenkinsfile.download_logs
```diff
@@ -55,7 +55,10 @@ pipeline {
         stage('Create tickets') {
             steps {
                 dir ('assisted-installer-deployment') {
-                    sh "skipper run ./tools/triage/create_triage_tickets.py --jira-access-token ${JIRA_ACCESS_TOKEN} -v"
+                    withCredentials([file(credentialsId: 'prod-pull-secret', variable: 'PULL_SECRET_FILE')]) {
+                      sh 'cp "${PULL_SECRET_FILE}" .'
```
osherdp, line 59: ok, I actually didn't think about the mounting part of this, which makes it a bit ugly IMHO (to copy it to the code dir)
it can also have some undesired affects if we're relying on ``git status/diff`` somewhere

in this case, I think ``PULL_SECRET`` (without specifying a file-path, just the value) might be the best solution here. this way it's easy to pass to the container and yet doesn't have the potential to leak in jenkins logs

javipolo: I'll see how that works out then

javipolo: Ok, now we pass the contents of the pull-secret to skipper via an env variable :+1:

osherdp (approved):

PR #389, opened 2023-06-29
Add signature for cases where the cluster moved from installing pending user action to error

eranco74:
We can probably close these tickets since the installation failed because the user didn't act and fixed the hardware issue (usually boot the hosts from the right disk) within the installation timeout

### DIFF
--- a/tools/triage/add_triage_signature.py
+++ b/tools/triage/add_triage_signature.py
@@ -973,6 +973,45 @@ class SlowImageDownload(ErrorSignature):
             self._update_triaging_ticket(report)
 
 
+class PendingUserAction(ErrorSignature):
+    cluster_status_regex = re.compile(r"Updated status of the cluster to (.*)")
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            **kwargs,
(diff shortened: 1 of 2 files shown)


tools/triage/add_triage_signature.py
```diff
@@ -970,6 +970,45 @@ def _process_ticket(self, url, issue_key):
...
+    def _process_ticket(self, url, issue_key):
+        md = get_metadata_json(url)
+
+        cluster = md["cluster"]
+
+        cluster_id = cluster["id"]
+        events = get_cluster_installation_events(url, cluster_id)
+        cluster_status_list = self._list_cluster_status_updates(events)
+        # This is signing triage tickets where the last status is always "error"
+        if cluster_status_list[-2] == "installing-pending-user-action":
```
nmagnezi, line 1003:
```suggestion
        if len(cluster_status_list) >= 2 and cluster_status_list[-2] == "installing-pending-user-action":
```

nmagnezi: This is just to avoid `IndexError: list index out of range`

nmagnezi (approved):
/lgtm

PR #390, opened 2023-06-29
NO-ISSUE: Fix DeletedHostsStatusSignature so it only triggers when needed

javipolo:
The signature was always triggering a message, with this it should only do when there are deleted_hosts


eranco74 (commented):
/lgtm

PR #393, opened 2023-07-18
Detect empty manifest bug

omertuc:
https://issues.redhat.com/issues/?jql=project%20in%20(12325045%2C%2012327120)%20AND%20cf%5B12317358%5D%20%3D%20SIGNATURE_empty_manifest


tools/triage/add_triage_signature.py
```diff
@@ -134,6 +134,17 @@ def get_metadata_json(cluster_url):
         raise FailedToGetMetadataException from e
 
 
+def get_manifests(cluster_url):
+    try:
+        res = requests.get("{}/cluster_files/manifests".format(cluster_url))
```
rccrdpccl, line 139: any reason why we choose format over f-strings?

omertuc: hmm just copied the metadata function

PR #407, opened 2023-08-28
MGMT-15569 - build script for move to appSRE

eifrach:
* Creating build script for migration to appSRE
* adding backward compatibility for the migration phase

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,20 @@
 .PHONY: all clean lint flake8 unit-test autopep8 isort pylint snapshot verify_snapshot tag_images release
 
+# backward compatibility 28-8-23
+ifdef container_command
+CONTAINER_CMD := $(container_command)
+endif
 
+
+CONTAINER_CMD := $(or $(CONTAINER_CMD), $(shell command -v podman 2> /dev/null))
+ifndef CONTAINER_CMD
+CONTAINER_CMD := docker
+endif
+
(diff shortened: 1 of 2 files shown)


eifrach:
> wait, didn't we talk about onboarding assisted-installer-cicd repo to appsre?
yes, so we are going to do it in 2 phases
1.  move all app's to appsre 
2.  move the repo / code to a assisted-installer-cicd 

after all moved to appsre moving the code will be much simpler, The code will also need some work/cleanup during the  move

osherdp:
> > wait, didn't we talk about onboarding assisted-installer-cicd repo to appsre?
> > yes, so we are going to do it in 2 phases
> 
> 1. move all app's to appsre
> 2. move the repo / code to a assisted-installer-cicd
> 
> after all moved to appsre moving the code will be much simpler, The code will also need some work/cleanup during the move

@rccrdpccl wdyt?
it seems to me like a total redundancy of components to introduce in app-interface

rccrdpccl:
> @rccrdpccl wdyt?
> it seems to me like a total redundancy of components to introduce in app-interface

What are we trying to achieve?
A lot of components of this repo are already migrated to assisted-installer-cicd

rccrdpccl:
@eifrach I think we can close this, right?

eifrach:
I will need it - but its already too old so i will recreate it

/close

PR #410, opened 2023-09-05
Set DEBUG as log level for clean-old-issues job

osherdp:
Makes sense to have more info rather than less info.

/cc @eliorerz @danmanor


eliorerz (approved):
LGTM

PR #411, opened 2023-09-10
NO-ISSUE: Increase the number of retries when experiencing difficulties connecting to the Jira server

danmanor:
Increase the number of retries when experiencing difficulties connecting to the `Jira` server.

### DIFF
--- a/tools/triage/clean_old_issues.py
+++ b/tools/triage/clean_old_issues.py
@@ -1,9 +1,10 @@
 import argparse
 import logging
 import os
-from typing import Final
+from typing import Callable, Final
 
 import jira
+from retry import retry
 
 from tools.jira_client import JiraClientFactory
 from tools.jira_client.consts import CLOSED_STATUS
@@ -51,7 +52,8 @@ def fetch_issues_from_jira(
         list of matching issues.
     """
 
-    result_list = jira_client.search_issues(
+    result_list = apply_function_with_retry(
+        jira_client.search_issues,
         f"project = {jira_project_key} AND attachments IS NOT EMPTY AND created <= -{created_before}d",
         maxResults=issues_limit,
     )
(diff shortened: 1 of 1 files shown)


tools/triage/clean_old_issues.py
```diff
@@ -60,6 +62,7 @@ def fetch_issues_from_jira(
     return list(result_list)
 
 
+@retry(exceptions=jira.JIRAError, tries=3, delay=10)
```
osherdp, line 65: it's retrying the whole process again from the start (for all issues), so I'm not sure it will get the affect we want

tools/triage/clean_old_issues.py
```diff
@@ -31,6 +32,7 @@ def get_logger(log_level: str) -> logging.Logger:
     return logger
 
 
+@retry(exceptions=jira.JIRAError, tries=3, delay=10)
```
osherdp, line 35: 10 seconds delay is a bit too much, I think we can do with 1 second

osherdp (changes requested):

tools/triage/clean_old_issues.py
osherdp: you also might try to re-close issues, I think it might raise some errors

tools/triage/clean_old_issues.py
```diff
@@ -106,6 +108,11 @@ def update_issues(jira_client: jira.JIRA, issues: list[jira.Issue], logger: logg
         raise ExceptionGroup("Failed to update all issues.", errors)
 
 
+@retry(exceptions=jira.JIRAError, tries=3, delay=1)
+def apply_function_with_retry(function: Callable, *args, **kwargs):
+    function(*args, **kwargs)
```
osherdp, line 113:
```suggestion
    return function(*args, **kwargs)
```
else ``result_list = apply_function_with_retry(...`` shouldn't work

osherdp (changes requested):

tools/triage/clean_old_issues.py
danmanor: my bad

osherdp:
@danmanor did you test it at our Jenkins instance?

danmanor:
@osherdp Yes I tested it now, looks ok - http://assisted-jenkins.usersys.redhat.com/job/test-cleanup-old-issues/2/console. I just added some logs to get which issues were successfully updated and which were not.

osherdp (approved):

PR #415, opened 2023-09-21
MGMT-13402: Using Jira attachments to produce ticket signatures

danmanor:
Currently, we use an `NFS` server, which requires `VPN` access to retrieve the data needed for producing the triage signatures. As we plan to transition our runtimes from Jenkins to an environment where `VPN` access might not be feasible, we want to adapt the system to get data from `Jira` attachments instead.

test pipelines :

- http://assisted-jenkins.usersys.redhat.com/job/test-download-logs/45/console  # with Jira
- http://assisted-jenkins.usersys.redhat.com/job/test-download-logs/46/console # with logs server

### DIFF
--- a/tests/tools/triage/utils/test_jira_remote_archive_querier.py
+++ b/tests/tools/triage/utils/test_jira_remote_archive_querier.py
@@ -1,0 +1,218 @@
+import io
+import tarfile
+import textwrap
+from unittest.mock import Mock
+
+import pytest
+
+from tools.triage.utils.extraction.jira_attachment_querier.jira_remote_archive_querier import (
+    JiraAttachmentsQuerier,
+)
+
+JIRA_ISSUE_KEY = "AITRIAGE-99999"
+
+
+@pytest.fixture
+def metadata_attachment() -> Mock:
+
+    metadata_attachment_mock = Mock()
+    metadata_attachment_mock.filename = "metadata.json"
+    metadata_attachment_mock.get.return_value = textwrap.dedent("""\
+        {
+            "name": "Foo Bar",
+            "age": "30",
+            "city": "New York",
+        }
+    """).encode()
+
+    return metadata_attachment_mock
+
+
+@pytest.fixture
+def cluster_files_attachment() -> Mock:
+
+    tar_buffer = io.BytesIO()
+
+    with tarfile.open(fileobj=tar_buffer, mode='w') as tar:
+        file1_content = io.BytesIO(b"Content of file1.txt")
+        file2_content = io.BytesIO(b"Content of file2.txt")
+
+        file1_info = tarfile.TarInfo(name='file1.txt')
+        file1_info.size = len(file1_content.getvalue())
+
+        file2_info = tarfile.TarInfo(name='file2.txt')
+        file2_info.size = len(file2_content.getvalue())
+
+        tar.addfile(file1_info, file1_content)
+        tar.addfile(file2_info, file2_content)
+
+    tar_buffer.seek(0)
+
+    cluster_files_attachment_mock = Mock()
+    cluster_files_attachment_mock.filename = "cluster_files.tar.gz"
+    cluster_files_attachment_mock.get.return_value = tar_buffer.getvalue()
+
+    return cluster_files_attachment_mock
+
+
+@pytest.fixture
+def jira_issue(metadata_attachment: Mock, cluster_files_attachment: Mock) -> Mock:
+    return Mock(key=JIRA_ISSUE_KEY, fields=Mock(attachment=[metadata_attachment, cluster_files_attachment]))
+
+
+@pytest.fixture(autouse=True)
+def jira_client(jira_issue: Mock) -> Mock:
+
+    def get_issue(issue_key: str) -> Mock:
+        match issue_key:
+            case "AITRIAGE-99999":
+                return jira_issue
+            case _:
+                return None
+
+    jira_client_mock = Mock()
+    jira_client_mock.issue.side_effect = get_issue
+    return jira_client_mock
+
+
+def test_init_with_existing_issue_should_be_successfull(jira_client):
+    querier = JiraAttachmentsQuerier(
+        jira_client,
+        JIRA_ISSUE_KEY,
+        "metadata.json",
+        Mock()
+    )
+
+    assert querier.downloaded
+    assert not querier.extracted
+
+
+def test_init_with_non_existing_issue_should_be_unsuccessfull(jira_client):
+    with pytest.raises(Exception):
+        JiraAttachmentsQuerier(
+            jira_client,
+            "non-existing-issue",
+            "metadata.json",
+            Mock()
+        )
+
+
+def test_init_download_with_existing_attachment_should_be_successfull(jira_client):
+    querier = JiraAttachmentsQuerier(
+        jira_client,
+        JIRA_ISSUE_KEY,
+        "metadata.json",
+        Mock()
+    )
+
+    assert querier.downloaded
+    assert not querier.extracted
+
+
+def test_init_download_with_non_existing_attachment_should_be_unsuccessfull(jira_client):
+    with pytest.raises(Exception):
+        JiraAttachmentsQuerier(
+            jira_client,
+            JIRA_ISSUE_KEY,
+            "non_existent_file.json",
+            Mock()
+        )
+
+
+def test_init_extraction_with_archive_path_should_be_successfull(jira_client):
+    querier = JiraAttachmentsQuerier(
+        jira_client,
+        JIRA_ISSUE_KEY,
+        "cluster_files.tar.gz",
+        Mock(),
+        True
+    )
+
+    assert querier.downloaded
+    assert querier.extracted
+
+
+def test_init_extraction_with_non_archive_path_should_be_unsuccessfull(jira_client):
+    querier = JiraAttachmentsQuerier(
+        jira_client,
+        JIRA_ISSUE_KEY,
+        "metadata.json",
+        Mock(),
+        True
+    )
+
+    assert querier.downloaded
+    assert not querier.extracted
+
+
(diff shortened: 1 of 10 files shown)


tools/triage/utils/extraction/jira_attachment_querier/jira_remote_archive_querier.py
```diff
@@ -0,0 +1,242 @@
...
+import nestedarchive
+
+from tools.triage.utils.extraction.jira_attachment_querier.exceptions import (
+    FailedToExtractAttachment,
+    FailedToGetAttachmentInIssue,
+)
+from tools.triage.utils.extraction.jira_attachment_querier.local_nested_archive_extractor import (
+    LocalNestedArchiveExtractor,
+)
+
```
danmanor, line 17: I created two options for getting the target file -
1. Extracting the whole archive using `LocalNestedArchiveExtractor` (for debugging purposes mainly as we can see the whole tree).
2. Extracting only necessary archives using `nestedarchive.get` .

danmanor:
@osherdp I think we can skip on using `pylint` on tests no ? it creates unnecessary comments

osherdp:
> @osherdp I think we can skip on using `pylint` on tests no ? it creates unnecessary comments

I think pylint is right here, it's wrong to use private members from outside of a class (even for unit-tests)

danmanor:
@osherdp , Where  there is a use of private members? I think `pylint` is complaining on the use of fixtures no? because it has the same name of the function

osherdp:
> @osherdp , Where there is a use of private members? I think `pylint` is complaining on the use of fixtures no? because it has the same name of the function

https://github.com/openshift-assisted/assisted-installer-deployment/pull/415/files#diff-af1904778f882a5283b31a479555aa9ac3c0a9d14f08bae18d7261259d7e18f2R106

danmanor:
@osherdp Oh I see, and in the case of fixtures (W0621), just tell pylint to ignore it ?

osherdp:
> @osherdp Oh I see, and in the case of fixtures (W0621), just tell pylint to ignore it ?

It somehow happens because of introducing ``tests/tools/triage/__init__.py`` and ``tests/tools/triage/utils/__init__.py``, without those files pylint passes

Be aware ``__init__.py`` files are no longer required starting Python 3.3. I'd vote to remove any redundant stuff

tools/triage/utils/extraction/jira_attachment_querier/jira_remote_archive_querier.py
```diff
@@ -0,0 +1,242 @@
...
+        new_path = target_directory / str(attachment.filename)
+
+        try:
+            with open(new_path, "wb") as file:
+                file.write(attachment_content)
+
+            return new_path
+
+        except Exception as e:
+            raise Exception(f"Failed to download the attachment to {new_path})") from e
```
osherdp, line 129: Both catching and raising ``Exception`` is wrong, as a general rule
In this specific case I would just remove it, us people know to read tracebacks and find where errors are coming from :)

osherdp: Also defining custom exceptions. Unless for the use-case of a library that wants its users to differentiate between different problems (e.g., ``requests`` that can return different errors if it's connection error, TLS problem, or something else), there's no need to have this

Let's apply those throughout this PR, we have a chance to remove a lot of code this way and to make it much easier to read and grasp

osherdp (changes requested):

danmanor:
@osherdp  https://github.com/openshift-assisted/assisted-installer-deployment/pull/415#issuecomment-1737794641
It still does not work for me without all __init__.py modules in tests/

Do you think we can disable these:
```       
redefined-outer-name,
unused-argument,
nvalid-name,
too-many-arguments,
too-few-public-methods
```
?

osherdp:
> @osherdp [#415 (comment)](https://github.com/openshift-assisted/assisted-installer-deployment/pull/415#issuecomment-1737794641) It still does not work for me without all **init**.py modules in tests/
> 
> Do you think we can disable these:
> 
> ```
> redefined-outer-name,
> unused-argument,
> invalid-name,
> too-many-arguments,
> too-few-public-methods
> ```
> 
> ?

No idea how you get it. For me:
```
$ git st
On branch change-sigs
Your branch is up to date with 'dan/change-sigs'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        deleted:    tests/tools/triage/__init__.py
        deleted:    tests/tools/triage/utils/__init__.py

no changes added to commit (use "git add" and/or "git commit -a")
[odp@osher-pc assisted-installer-deployment]$ skipper make pylint
...
pylint release/ tools/check_ai_images.py tools/jira_client tests/

------------------------------------
Your code has been rated at 10.00/10
```

tools/triage/utils/extraction/jira_attachment_querier/jira_remote_archive_querier.py
danmanor: @osherdp  We want to let the script run on all issues even if it failed to generate signatures for one of them, so we must catch some exceptions. Do you think we should raise all exceptions in the end ?

osherdp: I think we already do that on a per-issue basis (letting it run for the rest of the issues even if we encounter an error for one of them)

danmanor: I fixed the PR. I tried to avoid raising/catching/custom exceptions as much as possible, but I also wanted to make as less changes as possible to the code in order to avoid unnecessary issues, so I left the exception classes that was already there as some of the current code depends on it. With that being said, I also raises `FileNotFoundError` if the `jira` attachment download fails, in order to fit the behavior of `nestedarchive` in the case of fail

osherdp (changes requested):

tests/tools/triage/utils/test_jira_remote_archive_querier.py
```diff
@@ -0,0 +1,227 @@
...
+    jira_issue_mock.key = JIRA_ISSUE_KEY
+    return jira_issue_mock
+
+
+@pytest.fixture(autouse=True)
+def jira_client(jira_issue: MagicMock) -> MagicMock:  # pylint: disable=W0621
+
+    def get_issue(issue_key: str) -> MagicMock:
+        match issue_key:
+            case "AITRIAGE-99999":
```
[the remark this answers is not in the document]
danmanor: It is an interesting issue that I was also confused by, it seems that python `match` feature behaves unexpectedly in this case: there are two kind of patterns:

1. Value Pattern
2. Capture Pattern

In the second case - for example case y:
It doesn't compare against the value of y (if y was previously defined). Instead, it always matches any value and assigns that value to the variable y. It acts as "_" with assigning the value.

tools/triage/create_triage_tickets.py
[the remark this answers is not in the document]
danmanor: No, have you noticed any change I made that could effect it?

tools/triage/create_triage_tickets.py
```diff
@@ -137,6 +151,20 @@ def main(args):
...
+        help="whether to retrieve the data for signatures from the issue's attachments in the Jira server or from the logs server.",
+        required=False,
+    )
+    parser.add_argument("-d", "--dry-run", action="store_true", help="dry run")
+    parser.add_argument(
+        "--signatures_source",
+        default=os.environ.get("SIGNATURES_SOURCE"),
+        required=False,
+        help="pull secret acutal file contents",
+    )
```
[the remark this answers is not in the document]
danmanor: Can you elaborate the purpose ?

tests/tools/triage/utils/test_jira_remote_archive_querier.py
```diff
@@ -0,0 +1,218 @@
...
+                return jira_issue
+            case _:
+                return None
+
+    jira_client_mock = Mock()
+    jira_client_mock.issue.side_effect = get_issue
+    return jira_client_mock
+
+
+def test_init_with_existed_issue_should_be_successfull(jira_client):
```
rccrdpccl, line 78: nit: `s/existed/existing/g`

tools/triage/create_triage_tickets.py
```diff
@@ -53,18 +56,29 @@ def main(args):
...
+            failure.fields.summary.split()[2] if signatures_source == SignaturesSource.JIRA else failure["name"]
+        )
+        date = failure_name.split("_")[0]
         days_ago_creation = days_ago(date)
 
         if not args.all and days_ago_creation > DEFAULT_DAYS_TO_HANDLE:
             continue
 
-        logs_url = get_cluster_logs_base_url(failure["name"])
+        logs_url = get_cluster_logs_base_url(failure_name)
```
rccrdpccl, line 81: Why do we still need this? wouldn't this be unnecessary when JIRA signature source it's in use?

danmanor: @osherdp and I discussed it and figured it would be a good idea having both capabilities

tests/tools/triage/utils/test_jira_remote_archive_querier.py
danmanor: Thanks

tools/triage/create_triage_tickets.py
rccrdpccl: I meant having links pointing to the web server even when the file is coming from JIRA.
I agree that 2 modes are useful, but shouldn't we link the issue's uploaded file (or maybe nothing) instead of the nginx when the data comes from jira attachment?

danmanor: Oh got you. I missed that
fixing now thanks

PR #419, opened 2023-09-29
Look for multiple hosts holding the same VIP

omertuc:
This indicates keepalived issues / split brain, which might lead to
hosts trying to reach the API in servers aren't even running the
kube-apiserer

Example: https://issues.redhat.com/browse/AITRIAGE-8553?focusedId=23179977&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-23179977

### DIFF
--- a/tools/triage/add_triage_signature.py
+++ b/tools/triage/add_triage_signature.py
@@ -3018,6 +3018,68 @@ class MissingMC(ErrorSignature):
             self._update_triaging_ticket("Missing rendered MachineConfig issue detected")
 
 
+class DuplicateVIP(ErrorSignature):
+    """
+    Looks for nodes holding the same VIP
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            **kwargs,
+            comment_identifying_string="h1. VIP found in multiple nodes",
+            function_impact_label="vip_duplicate",
+        )
+
+    def _process_ticket(self, url, issue_key):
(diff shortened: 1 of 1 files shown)


tools/triage/add_triage_signature.py
```diff
@@ -2983,6 +2983,68 @@ def _process_ticket(self, url, issue_key):
...
+
+    def _process_ticket(self, url, issue_key):
+        cluster = get_metadata_json(url)["cluster"]
+
+        # SNO doesn't need balancing
+        if cluster.get("high_availability_mode") == "None":
+            return
+
+        # VIPs are not relevant with load balancer
+        if cluster.get("user_managed_networking") == True:
```
rccrdpccl, line 3007:
```suggestion
        if cluster.get("user_managed_networking"):
```

omertuc: If `user_managed_networking` has any value other than `True` (e.g., `5`) then I prefer this `if` to not have its body executed, so your diff suggestion isn't technically equivalent.

The concept of values having "truthiness" is in my opinion a very unfortunate unnecessary complication that sadly most programming languages have which I prefer to not rely on.

rccrdpccl: I agree, but linter disagrees :)

rccrdpccl: We could ignore linting though

